PR #3, opened 2018-04-03
Added custom options support

delvedor:
As titled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,16 @@ module.exports = function (fastify, opts, next) {
 // optional
 module.exports.autoload = false
 ```
+
+If you want to pass some custom options to the registered plugins via `fastify-autoload`, use the `options` key:
+```js
+fastify.register(AutoLoad, {
+  dir: path.join(__dirname, 'foo'),
+  options: { foo: 'bar' }
+})
+```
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -53,6 +53,14 @@ module.exports = function (fastify, opts, next) {
...
 module.exports.autoload = false
 ```
+
+If you want to pass some custom options to the registered plugins via `fastify-autoload`, use the `options` key:
+```js
+fastify.register(AutoLoad, {
+  dir: path.join(__dirname, 'foo'),
+  options: { foo: 'bar' }
+})
+```
```
mcollina: are the options shared by all plugins? Are they prefixed?

delvedor: Yes, it is passed to all the auto registered plugins.
If user want different options based on the plugin it can prefix the options.
```
fastify.register(AutoLoad, {
  dir: path.join(__dirname, 'foo'),
  options: {
    auth: {},
    endpoint: {},
  }
})
```

Since internally we are using `Object.assign` the prefix property it is not lost.

mcollina: Can you make this explicit? "options will be passed to all loaded plugins".

delvedor: Done :)

PR #7, opened 2018-05-19
 append plugin autoPrefix to default prefix

thekemkid:
Using `prefix` in the default autoload options should apply a prefix for all plugins loaded by this module. This makes it so any modules then loaded by autoload appends the modules autoPrefix to the end of the default autoload prefix.

It also adds a prefixOverride to override the defaultPrefix (and autoPrefix).

I hope this is okay!

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,68 @@ fastify.register(AutoLoad, {
 ```
 *Note that options will be passed to all loaded plugins.*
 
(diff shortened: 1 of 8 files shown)


thekemkid:
Note: this breaks previous behaviour. If both a default prefix option got set and autoPrefix was set, it would default to the default prefix.

mcollina:
cc @delvedor

mcollina:
Thanks @thekemkid!

Issue #9, opened 2018-08-15
Autoload ignore/include pattern


peteyycz:
I like to keep my tests in the same folder as the original files.

```
services/healthz.js
services/healthz.test.js
```

Is it possible to ignore `*.test.js` files from autoloading?

peteyycz:
I'd like to work on it, yes! :)

mcollina:
Fixed!

PR #10, opened 2018-08-15
Autoload ignore pattern

peteyycz:
Add ignore pattern to autoload configuration to ignore specific file in the autoload folder.

Related issue: #9


peteyycz:
How does the release process looks like? When should I expect this to be released? :)

PR #19, opened 2018-12-19
Skip loading folders which have no .js files inside

arniu:


arniu:
I'm really sorry for that. The original code was a bit of hard to read, so I ...

Now, what can I do to revert it?

mcollina:
I would recommend to start from scratch and porting over your changes. Thanks!

arniu:
OK. Wait a minute please～

PR #20, opened 2018-12-19
Skip loading folders without .js files inside

arniu:


mcollina:
Can you add a unit test?

arniu:
Done!

PR #21, opened 2019-02-20
Typescript support

Rubendewit:
Files with the Typescript extension were skipped, thereby making this project unusable in Typescript projects.

### DIFF
--- a/index.js
+++ b/index.js
@@ -34,15 +34,15 @@ module.exports = function (fastify, opts, next) {
             }
 
             cb(null, {
-              // skip directories withoud .js files inside
-              skip: files.every(name => !name.match(/.js$/)),
+              // skip directories without .js or .ts files inside
+              skip: files.every(name => !name.match(/.(ts|js)$/)),
               file: toLoad
             })
           })
         } else {
           cb(null, {
-            // only accept .js files
-            skip: !(stat.isFile() && file.match(/.js$/)),
+            // only accept .js and .ts files
+            skip: !(stat.isFile() && file.match(/.(ts|js)$/)),
             file: toLoad
           })
         }
--- a/test/basic.js
+++ b/test/basic.js
@@ -3,7 +3,7 @@
 const t = require('tap')
 const Fastify = require('fastify')
 
-t.plan(32)
+t.plan(35)
 
 const app = Fastify()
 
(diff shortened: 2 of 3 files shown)


mcollina:
Node.js cannot load `ts` file on its own. Can you add a full example that is actually using typescript? Normally those are transpiled and then loaded, so it should not be an issue.
Maybe you are working on a system using `ts-node`?

Rubendewit:
I am indeed using `ts-node`. Updating this package for the sole purpose of allowing `ts` files seems a bit too much, if I could refactor a bit and remove `ts-node` from our project. My apologies!

trixden:
Why this PR did closed? Will this library ever support Typescript?

Rubendewit:
> Why this PR did closed? Will this library ever support Typescript?

I closed the PR, because I worked around the issue I was encountering. As mcollina said, .ts files would be transpiled and then loaded. While developing, however, I use ts-node, which means we still try to load the .ts files.

I circumvented this issue by ditching ts-node, and doing a `tsc & run build` setup. This means the files loaded in by `fastify-autoload` aren't .ts files, but regular .js files.

trixden:
> I closed the PR, because I worked around the issue I was encountering. As mcollina said, .ts files would be transpiled and then loaded. While developing, however, I use ts-node, which means we still try to load the .ts files.
> 
> I circumvented this issue by ditching ts-node, and doing a `tsc & run build` setup. This means the files loaded in by `fastify-autoload` aren't .ts files, but regular .js files.

Ohh. I understand. Thanks. I think, i write local library based on this fork, and dirs prefix i will pass from environment configuration file. That's will be perfect.

despian:
Can this possibly be re-opened and merged?

I think it's useful to use ts-node in dev. It's much better when using nodemon for example, saves re-building every time you change a file. Also, it means errors point to the ts source rather than the compiled js.

mcollina:
Possibly, would you like to revamp this PR?
I'm happy for something similar to land as long as the `ts` wildcard is used only if `ts-node` is used.

despian:
Sure, happy to submit a new PR.

Though, rather than checking for ts-node explicitly how about just having an option to include ts files? That way it could support other setups which may what to load ts files.

mcollina:
Sure.

mcollina:
+1

Il giorno ven 14 giu 2019 alle 18:34 despian <notifications@github.com> ha
scritto:

PR #23, opened 2019-03-16
check for exports.default first

ekoeryanto:


mcollina:
Can you please add a unit test?

PR #26, opened 2019-04-09
autoload support commonJS that export default XX (es6) transfer to commonjs

kidyfirst:
solve 25#
and add types file


Eomm:
Is this PR the same subject of #23 ?

kidyfirst:
ok

PR #28, opened 2019-04-11
support types and add types case

kidyfirst:


package.json
```diff
@@ -28,13 +30,16 @@
...
+    "http": "^0.0.0",
     "pre-commit": "^1.2.2",
     "snazzy": "^8.0.0",
     "standard": "^12.0.1",
-    "tap": "^12.5.3"
+    "tap": "^12.5.3",
+    "typescript": "^3.4.3"
   },
   "dependencies": {
+    "@types/node": "^11.13.2",
```
mcollina: Can you move this to devDependencies?

package.json
```diff
@@ -28,13 +30,16 @@
   },
   "homepage": "https://github.com/fastify/fastify-autoload#readme",
   "devDependencies": {
-    "fastify": "^2.0.0",
+    "fastify": "^2.2.0",
+    "http": "^0.0.0",
```
mcollina: this is not a thing, remove it

mcollina:
@Ethan-Arrowood @delvedor would you mind to take a look?

package.json
kidyfirst: ok

package.json
kidyfirst: have removed

PR #30, opened 2019-04-11
support loading of plugins in a smart order based on dependency

### LINKED ISSUE #29: feature: load plugins in order based on dependencies
Say we have `plugin-a` & `plugin-b` in `./plugins` directory. And `plugin-b` is added as a  `dependencies` in `plugin-a`.

With this an error is thrown, because `plugin-a` is getting loaded before the `plugin-b`. This happens due to the plugin files being read in the alphabetic order.

It would be good if `fastify-autoload` is smart enough to load the plugins in order they're being used.

salmanm:
Fixes #29 

- Auto loads plugins in the order they are needed (checks for the `dependencies` array).
- Checks for cyclic dependencies and returns error, if found.


Eomm:
Hi,
thanks for the PR!

Could you [fix the error](https://travis-ci.org/fastify/fastify-autoload/jobs/518795928) for Node.js 6?
It should be necessary replace the `Object.value` with a `Object.keys()` and then grab the value.

Meanwhile, I'll check deeper the code, for now, I'm thinking that a hidden limit is that it will be not possible load twice a plugin with the same name but I need to think if it is a real use case

index.js
```diff
@@ -63,27 +76,64 @@ module.exports = function (fastify, opts, next) {
...
             pluginOptions.prefix = prefix + plugin.autoPrefix
           }
+
           if (plugin.prefixOverride !== void 0) {
             pluginOptions.prefix = plugin.prefixOverride
           }
 
-          if (plugin.autoload !== false) {
-            fastify.register(plugin.default || plugin, pluginOptions)
+          allPlugins[pluginName] = {
```
mcollina: I think we should throw if two plugins have the same name.z

salmanm:
Thanks @mcollina, done those changes.

salmanm:
@Eomm I tried to think about the hidden limit you mentioned. Loading a plugin twice in the same load chain is a result of cyclic dependency. I couldn't think of a case where one would load a plugin twice in the **same load chain** and still not cause cyclic dependency.

It's however fine to load a plugin from two separate load sequence. i.e. `plugin-a` & `plugin-b` both needs `plugin-c`.

mcollina:
Also, CI is failing.

salmanm:
Yep, good idea.
I didn't do it earlier as I thought fastify already blocks when `dependencies` aren't loaded already.
But no harm checking.

All done now.

index.js
```diff
@@ -63,27 +76,69 @@ module.exports = function (fastify, opts, next) {
...
           }
+
           if (plugin.prefixOverride !== void 0) {
             pluginOptions.prefix = plugin.prefixOverride
           }
 
-          if (plugin.autoload !== false) {
-            fastify.register(plugin.default || plugin, pluginOptions)
+          if (allPlugins[pluginName]) {
+            throw new Error('Duplicate plugin name')
```
Eomm:
```suggestion
            throw new Error(`Duplicate plugin: ${pluginName}`)
```

index.js
```diff
@@ -63,27 +76,69 @@ module.exports = function (fastify, opts, next) {
...
+
+        registerPlugin(name, plugin, options)
+      }
+
+      const pluginKeys = Object.keys(allPlugins)
+      for (let i = 0; i < pluginKeys.length; i++) {
+        cyclicDependencyCheck = {}
+
+        try {
+          const p = allPlugins[pluginKeys[i]]
```
Eomm: This `p` var could be removed (GC min benefit)

salmanm:
@Eomm Done

PR #33, opened 2019-04-22
Load multiple modules from a directory

PatrickHeneise:
This feature allows building routes from files automatically and prefixes the routes with the folder name.

As an example, a typical API structure:

- /services/products/list.js
- /services/products/get.js

both files have similar code:

list:

```
module.exports = function(fastify, opts, next) {
  fastify.route({
    method: 'GET',
    url: '/',
    handler: function(request, reply) {
      reply.send({ hello: 'world' })
    }
  })
  next()
}
```

get:

```
module.exports = function(fastify, opts, next) {
  fastify.route({
    method: 'GET',
    url: '/:id',
    handler: function(request, reply) {
      reply.send({ hello: 'world' })
    }
  })
  next()
}
```

Without this feature, I'd need a separate index.js file with this content:

```
module.exports = function(fastify, opts, next) {
  fastify.register(require('./list'))
  fastify.register(require('./get'))
  next()
}
```

To load each module. With this patch, routes will automatically be loaded as `localhost:3000/products` and `localhost:3000/products/:id`

There are some complications with that approach, if `a.js` uses `/:id` and `b.js` uses `/something`, `/something` is not reachable because it's loaded after a.js. I'll address those in a separate issue.

Let me know what you think and if it's good to merge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,6 +137,40 @@ fastify.register(AutoLoad, {
 })
 ```
 
+fastify-autoload loads folders with route definitions automatically, without explicitly registering them. The folder name is used as default prefix for all files in that folder, unless otherwise specified in an `index.js`. See "module.exports.autoPrefix" on how to overwrite this behaviour.
+
+```js
+// index.js
+fastify.register(AutoLoad, {
+  dir: path.join(__dirname, 'services'),
+  options: {}
+})
+
+// /services/items/get.js
+module.exports = function (f, opts, next) {
+  f.get('/:id', (request, reply) => {
+    reply.send({ answer: 42 })
+  })
+
+  next()
+}
+
+// /services/items/list.js
+module.exports = function (f, opts, next) {
+  f.get('/', (request, reply) => {
+    reply.send([0, 1, 2])
+  })
+
+  next()
+}
+
+/**
+ * Routes generated:
+ * GET /items
+ * GET /items/:id
+ */
+```
+
 ## License
 
 MIT
--- a/index.js
+++ b/index.js
@@ -53,6 +53,12 @@ module.exports = function (fastify, opts, next) {
               for (let index = 0; index < files.length; index++) {
                 const file = files[index]
 
+                // windows debugging
+                console.log(file)
+                console.log(toLoad)
+                console.log(toLoad.split('/').pop())
+                // end windows debugging
+
                 plugins.push({
                   skip: !file.match(/.js$/),
                   opts: {
--- a/index.js
+++ b/index.js
@@ -47,6 +47,7 @@ module.exports = function (fastify, opts, next) {
             // if the directory contains files but no index.js, load them as independend plugins
             if (
               files.indexOf('index.js') === -1 &&
+              files.indexOf('package.json') === -1 &&
               files.toString().indexOf('.js') > -1
             ) {
               let plugins = []
--- a/index.js
+++ b/index.js
@@ -44,11 +44,28 @@ module.exports = function (fastify, opts, next) {
               return
             }
 
-            cb(null, {
-              // skip directories withoud .js files inside
-              skip: files.every(name => !name.match(/.js$/)),
-              file: toLoad
-            })
+            // if the directory contains files but no index.js, load them as independend plugins
+            if (
+              files.indexOf('index.js') === -1 &&
+              files.toString().indexOf('.js') > -1
+            ) {
+              let plugins = []
+              for (let index = 0; index < files.length; index++) {
+                const file = files[index]
+
+                plugins.push({
+                  skip: !file.match(/.js$/),
+                  file: path.join(toLoad, file)
+                })
+              }
+              cb(null, plugins)
+            } else {
+              cb(null, {
+                // skip directories without .js files inside
+                skip: files.every(name => !name.match(/.js$/)),
+                file: toLoad
+              })
+            }
           })
         } else {
           cb(null, {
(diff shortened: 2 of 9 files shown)


PatrickHeneise:
As far as I understand, there was no prefix when `index.js` was present before, so there is no change in behaviour. The prefix can be defined in `index.js`:

```
module.exports = function(fastify, opts, next) {
   fastify.register(require('./list'))
   fastify.register(require('./get'))

    next()
 }

 module.exports.autoPrefix = '/something'
```

works in this version the same way as just storing get/list in /something.

What I don't understand is how `answer.js` was loaded in the tests in the first place. Does `main` in the package.json compensate for an index.js?

PatrickHeneise:
I added test cases with and without `index.js` :)

mcollina:
> What I don't understand is how answer.js was loaded in the tests in the first place. Does main in the package.json compensate for an index.js?

Yes exactly. If you specify a `main` field, that file will be loaded in place of the `index.js` when you `require('./folder')`.

PatrickHeneise:
Of course not. Will do

index.js
```diff
@@ -58,16 +78,18 @@ module.exports = function (fastify, opts, next) {
...
         }
       })
-    }, (err, stats) => {
+    }, (err, files) => {
       if (err) {
         next(err)
         return
       }
 
+      const stats = [].concat(...files)
```
delvedor: Probably I'm missing something, what's the need of this line?

PatrickHeneise: `files` can contain further arrays if there are multiple files in a folder:

```js
[ { skip: true,
    file: '/api/services/.DS_Store' },
  [ { skip: false,
      opts: [Object],
      file: '/api/services/orders/get.js' },
    { skip: false,
      opts: [Object],
      file: '/api/services/orders/list.js' } ] ]
```

`[].concat(...files)` flattens the array.

mcollina:
Is the behavior re-entrant? What happens if we have a path/to/something.js file?

PatrickHeneise:
autoload isn't recursive. If `/services` is the entry point, `/services/path/to/something.js` is skipped, as `/path/to` does not contain any .js files. There's a different library to load recursively and add more functionality as well: https://github.com/fastify/help/issues/76#issuecomment-485403567

README.md
```diff
@@ -137,6 +137,40 @@ fastify.register(AutoLoad, {
...
 ```
 
+fastify-autoload loads folders with route definitions automatically, without explicitly registering them. The folder name is used as default prefix for all files in that folder, unless otherwise specified in an `index.js`. See "module.exports.autoPrefix" on how to overwrite this behaviour.
+
+```js
+// index.js
+fastify.register(AutoLoad, {
+  dir: path.join(__dirname, 'services'),
+  options: {}
+})
```
salmanm: Can we add an option to disable auto route prefixing? Or rather keep it off by default and enable it here with a flag say `autoRoutePrefix: true`?

PatrickHeneise: It's automatically disabled when an `index.js` is present in the folder..

salmanm: Not sure if that's explicit enough.

PatrickHeneise: I think it is quite explicit. The code is only executed:

- if there is a subfolder
- if the folder contains .js files
- if the folder does not contain an index.js file with further configuration
- if the file registers a route (fastify-plugin doesn't use prefix)

I've tested a bunch of existing use cases in my project and the current behaviour doesn't change. Auto-prefixing routes makes total sense to me, as it reduces the amount of additional configuration. If absolutely needed, I would enable it be default and add `autoRoutePrefix: false` to the options when to turn the feature off.

What do you think?

PatrickHeneise:
Please do not merge. There are some path issues with Windows I didn't see before.

PatrickHeneise:
@mcollina I added a test for `package.json` to exclude the case where a directory is auto-prefixed when a package.json is present, so this is not a breaking change any longer.

With the current version of autoload, this specific scenario returns:

`UnhandledPromiseRejectionWarning: Error: Cannot find module`

Thanks everyone for the guidance and feedback! I hope this is now ready to merge :)

mcollina:
Can you please articulate that unhandledRejection case? It seems something bad that we should indeed fix: catching that rejection and forwarding it up the chain.

mcollina:
On second thought, I would also like to add the flag. If you check the template generated by fastify-cli, you can see there is a “plugins” folder. IMHO we should not automatically apply the prefix there.

What do you think of this case?

PatrickHeneise:
"Be aware that if you use fastify-plugin this option won't work." - https://www.fastify.io/docs/latest/Routes/#route-prefixing

plugins are not prefixed, shouldn't be an issue.

PatrickHeneise:
> Can you please articulate that unhandledRejection case? It seems something bad that we should indeed fix: catching that rejection and forwarding it up the chain.

I opened a separate issue for the rejection error: https://github.com/fastify/fastify-autoload/issues/34

PatrickHeneise:
> On second thought, I would also like to add the flag. If you check the template generated by fastify-cli, you can see there is a “plugins” folder. IMHO we should not automatically apply the prefix there.
> 
> What do you think of this case?

autoprefix is _only_ applied to folders and if there is no index.js nor package.json present in the folder, which is currently a must, otherwise there are unhandled rejection errors for routes and plugins alike. That means this scenario can't exist right now.

I'd like to keep things simple and follow convention over configuration. Nuxt.js does the same thing:

> Nuxt.js automatically generates the vue-router configuration based on your file tree of Vue files inside the pages directory." - https://nuxtjs.org/guide/routing

If someone wants to disable auto-prefixing, they likely need additional configuration. For that, they'll need to add an `index.js` in the folder - which automatically disabled the autoprefix.

mcollina:
Ok, let’s release this!

PatrickHeneise:
If my assumptions are wrong on this, I'll add the config options and default autoprefix to false.

Thanks for merging it!

PR #41, opened 2019-06-16
ts-node support

despian:
Following the discussion in (#21) this is a revamped PR to support loading plugins from `.ts` files directly when using `ts-node`.

A new option has been added `includeTypeScript` which, when set to `true`, will allow loading of both `.js` and `.ts` files.


despian:
Absolutely, the 4 space indentation was triggering me a little too 😄 I was just trying to stick to the patterns already used in the existing ts files.

However, ts linting with `standard` does not work. There are a whole host of issues on the `standrd` tracker about it. e.g. [standard#1283](https://github.com/standard/standard/issues/1283)

Its usage in the main `fastify` project works somewhat since some `eslint` exception comments are added to the top of the ts files. But these are not ideal since they just disable certain checks.

The [eslint-config-standard-with-typescript](https://github.com/mightyiam/eslint-config-standard-with-typescript) project works well for linting ts and is compatible with the `standard` ruleset. I suggest using this for linting ts until there is better support in `standard`.

despian:
Looks like the use of the spread syntax ('...') in `eslint-plugin-node` breaks compatibility with node 6.

I'll see if I can do something about that.

despian:
So the issue was `eslint-config-standard-with-typescript` has a newer version of `eslint-plugin-node` in its peer dependencies which is incompatible with node 6.

I just removed its peers from the devDependcies since they are all provided as dependencies of `standard` anyhow. However, this does mean you get a couple of warnings in about missings peers when running `npm install` since the versions specified in `eslint-config-standard-with-typescript` are newer. This is not ideal but everything does work. I don't think the newer versions are actually required.

despian:
@mcollina now this has some tests running with `ts-node` it seems to me that having the file [types.test.ts](https://github.com/fastify/fastify-autoload/blob/23019ab403d48a56ee3328675f3e6280db6b5bb1/types.test.ts) and running `tsc` against it in [package.json#L9](https://github.com/fastify/fastify-autoload/blob/23019ab403d48a56ee3328675f3e6280db6b5bb1/package.json#L9) is redundant.

I could add a commit with some more specific tests against the option types and remove the unneeded `types.test` file. What do you think?

The big advantage of doing the type testing with `ts-node` rather than simply using `tsc` is that the code is compiled and executed rather than just compiled. So it ensures the defined types actually match up with the underlying `js` code.

Having type tests run in this way would have caught an error I came across in the types of another plugin [fastify-sensible#20](https://github.com/fastify/fastify-sensible/pull/20). Where the type for an object was incorrectly exported with the wrong name. There were type tests for this but they were also using the wrong name. And since the type tests are never actually executed it was not caught.

Maybe it's worth considering moving other type tests within @fastify to use `ts-node`. Or even to just convert the normal tests to `ts` and run them all with ts-node. If I were to submit some more PRs to other plugins along these lines do you think they would be accepted?

despian:
You could do something like:

```js
const moduleList = Object.keys(require.cache).join('\n')
const isTsNode = /ts-node\/dist\/index\.js$/m.test(moduleList)
```

This code seems a little brittle to me though and likely to require maintenance. Also, it doesn't allow for other scenarios where people might want to load `.ts` files. E.g. what if someone creates some fork `super-ts-node` and wants to use that.

Seemed to make more sense to me to just have an explicit user definable option for loading `.ts` files.

Though maybe we could add this check to load `.ts` automatically but also have the explicit option to force loading of `.ts`. Would be 0 config in most cases but also allows for other scenarios where this check isn't sufficient.

Ethan-Arrowood:
I agree with this change. It supports two kinds of TS dev styles (with and without ts-node). I rarely use ts-node. I almost always just run the compiled js, but there are many great perks from using `ts-node` so I'm glad it is seeing some support here. 

@despian i'm close to finishing a revamp of the core fastify types. Once that lands I'd be interested in seeing what is possible with `ts-node`. For what it is worth, Fastify prioritizes JavaScript development and then works to secondarily provide TypeScript support. 

--- 

There is this new library I've been a big fan of for testing type definitions: https://github.com/SamVerschueren/tsd 

I've looked at ts-node in the past but never considered it for this use case. I'm always happy to see alternative type definition methods.

Issue #35, opened 2019-04-25
mjs compatibility in Node 12


mcollina:
Node 12 has landed EcmaScript Modules support in a shape that would resemble the final result. IMHO it’s time to test this module and make sure that we can load native .mjs files.

We might need to refactor certain parts, but I think it’s worthwhile - we can also use this to open issue against Node core and provide feedbacks on modules themselves.

PR #52, opened 2019-08-30
Add support for .mjs files

SerayaEryn:
This add support to load `.mjs` files as suggested in #35 

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -14,12 +14,12 @@ module.exports = function (fastify, opts) {
 
     const allPlugins = {}
     const promises = []
-    for (const { skip, file, opts, isModule } of stats) {
+    for (const { skip, file, opts } of stats) {
       if (skip) {
         continue
       }
 
-      promises.push(importPluginAndMetaData(file, opts, isModule, allPlugins, defaultPluginOptions))
+      promises.push(importPluginAndMetaData(file, opts, allPlugins, defaultPluginOptions))
     }
     return Promise.all(promises)
       .then(() => allPlugins)
@@ -67,8 +67,8 @@ module.exports = function (fastify, opts) {
 // same as fastify-plugin
 module.exports[Symbol.for('skip-override')] = true
 
-function importPluginAndMetaData (file, opts, isModule, allPlugins, defaultPluginOptions) {
-  return importPlugin(file, isModule).then((content) => {
+function importPluginAndMetaData (file, opts, allPlugins, defaultPluginOptions) {
+  return importPlugin(file).then((content) => {
     const plugin = wrapRoutes(content)
     const pluginConfig = (plugin.default && plugin.default.autoConfig) || plugin.autoConfig || {}
     const pluginOptions = Object.assign({}, pluginConfig, defaultPluginOptions)
@@ -128,26 +128,30 @@ function wrapRoutes (content) {
   }
 }
 
-function importPlugin (file, isModule) {
-  if (isModule) {
-    return import(file)
-  } else {
+function importPlugin (file) {
+  if (file.endsWith('.js') || file.endsWith('.cjs')) {
     try {
-      return Promise.resolve(require(file))
-    } catch (err) {
-      return Promise.reject(err)
+      const module = require(file)
+      return Promise.resolve(module)
+    } catch (error) {
+      if (error.code === 'ERR_REQUIRE_ESM') {
+        return import(file)
+      }
+      return Promise.reject(error)
     }
+  } else {
+    return import(file)
   }
 }
 
 function loadPluginFiles (opts) {
   const packagePattern = /^package\.json$/im
   const indexPattern = opts.includeTypeScript
-    ? /^index\.(ts|js|mjs)$/im
-    : /^index\.(js|mjs)$/im
+    ? /^index\.(ts|js|mjs|cjs)$/im
+    : /^index\.(js|mjs|cjs)$/im
   const scriptPattern = opts.includeTypeScript
-    ? /((^.?|\.[^d]|[^.]d|[^.][^d])\.ts|\.js|\.mjs)$/im // For .ts files, ignore .d.ts
-    : /\.(js|mjs)$/im
+    ? /((^.?|\.[^d]|[^.]d|[^.][^d])\.ts|\.js|\.mjs|\.cjs)$/im // For .ts files, ignore .d.ts
+    : /\.(js|mjs|cjs)$/im
 
   return readDir(opts.dir).then((list) => {
     const promises = []
@@ -171,8 +175,7 @@ function loadPluginFile (opts, file, packagePattern, indexPattern, scriptPattern
       return [{
         // only accept script files
         skip: !(stat.isFile() && scriptPattern.test(file)),
-        file: toLoad,
-        isModule: file.endsWith('.mjs')
+        file: toLoad
       }]
     }
   })
@@ -181,7 +184,6 @@ function loadPluginFile (opts, file, packagePattern, indexPattern, scriptPattern
 function findPluginsInDirectory (toLoad, packagePattern, indexPattern, scriptPattern) {
   return readDir(toLoad).then((files) => {
     const fileList = files.join('\n')
-    const moduleDefinedByPackageJson = hasModuleDefinedInPackageJson(toLoad)
     // if the directory does not contain a package.json or an index,
     // load each script file as an independend plugin
     if (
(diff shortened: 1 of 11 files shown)


index.js
```diff
@@ -2,188 +2,202 @@
 
 const fs = require('fs')
 const path = require('path')
-const steed = require('steed')
 
-module.exports = function (fastify, opts, next) {
+module.exports = async function (fastify, opts) {
+  await loadAllPlugins(fastify, opts)
+}
```
mcollina: Unfortunately this still support Node.js 6. We can move to async/await it when we cut Fastify v3.

test/mjs/plugins/plugin.mjs
```diff
@@ -0,0 +1,9 @@
+'use strict'
+
+export default function pluginMjs (f, opts, next) {
+  f.get('/mjs', (request, reply) => {
+    reply.send({ hello: 'world' })
+  })
+
+  next()
+}
```
mcollina: It'd be good to add a test with a module using fastify-plugin

SerayaEryn: I‘ll do that.

index.js
SerayaEryn: I could refactor it to use Promises only.

Or we do not merge this until fastify v3 has been cut.

mcollina: @delvedor what do you think?

index.js
```diff
@@ -2,188 +2,202 @@
...
+  return new Promise((resolve, reject) => {
+    fs.stat(toLoad, (err, stat) => {
+      if (err) {
+        reject(err)
+      } else {
+        resolve(stat)
       }
+    })
+  })
+}
```
mcollina: Can you please use `util.promisify` instead of manually wrapping them?

index.js
```diff
@@ -2,188 +2,202 @@
 
 const fs = require('fs')
 const path = require('path')
-const steed = require('steed')
```
mcollina: should we remove this from the dependencies then?

index.js
```diff
@@ -2,188 +2,202 @@
...
-  }
+  const stats = [].concat(...files)
+
+  const allPlugins = {}
 
-  fs.readdir(opts.dir, function (err, list) {
-    if (err) {
-      next(err)
-      return
+  for (let i = 0; i < stats.length; i++) {
```
mcollina: you can use a for-of loop here instead.

index.js
delvedor: Fastify v3 will be out by the end of the year (more or less).
I'm not using es modules so I don't see the urge to do this, but it may be useful to support them.
I'm fine with both options.

mcollina: Let's go with promises, I prefer this to be shipped asap so we can test it out.

index.js
SerayaEryn: I used `es6-promisify` for now since node 6 has no `util.promisify`: 
https://nodejs.org/docs/latest-v6.x/api/util.html

index.js
SerayaEryn: I removed it.

index.js
SerayaEryn: I replaced it with a for-of loop.

index.js
```diff
@@ -2,149 +2,75 @@
...
-          return
-        }
-      }
+            allPlugins[pluginName] = {
+              plugin,
+              name: pluginName,
+              dependencies: pluginMeta.dependencies,
+              options: pluginOptions
+            }
+          })
```
mcollina: Would you mind moving the above in its own function?

SerayaEryn: I moved it to a function.

index.js
```diff
@@ -172,18 +53,154 @@ module.exports = function (fastify, opts, next) {
...
+    })
+}
+
+function loadPluginFile (opts, file, packagePattern, indexPattern, scriptPattern) {
+  if (opts.ignorePattern && file.match(opts.ignorePattern)) {
+    return Promise.resolve([{ skip: true }])
+  }
+  const toLoad = path.join(opts.dir, file)
+  return getStat(toLoad)
+    .then((stat) => {
```
mcollina: In general, I prefer to have these two in the same line `getStat(toLoad).then((stat) => {`. This will limit nesting and make call a bit easier to read.

index.js
```diff
@@ -172,18 +53,154 @@ module.exports = function (fastify, opts, next) {
...
+              }
+              return plugins
+            } else {
+              return [{
+                // skip directories without script files inside
+                skip: files.every(name => !scriptPattern.test(name)),
+                file: toLoad
+              }]
+            }
+          })
```
mcollina: Can you extract this as well?

package.json
```diff
@@ -8,7 +8,7 @@
     "lint": "standard | snazzy && npm run lint:typescript",
     "lint:fix": "standard --fix",
     "lint:typescript": "standard --parser @typescript-eslint/parser --plugin @typescript-eslint/eslint-plugin ./*.ts",
-    "test": "standard | snazzy && tap test/*.js && npm run typescript",
+    "test": "NODE_OPTIONS=--experimental-modules standard | snazzy && tap test/*.js && npm run typescript",
```
mcollina:
```suggestion
    "test": "NODE_OPTIONS=--experimental-modules standard | snazzy && tap --no-esm test/*.js && npm run typescript",
```

SerayaEryn: What is the reason for this change?
With this change 5 tests would fail.

index.js
SerayaEryn: I extracted it.

index.js
SerayaEryn: I changed that.

SerayaEryn:
> I think this needs to support https://nodejs.org/api/esm.html#esm_code_package_json_code_code_type_code_field, both at the top level and inside folders.
I'll take a look at that.

package.json
mcollina: The reason is that you are not targeting the Node experimental modules, but rather the `esm` implementation. They are going to be significantly different.

test/mjs/plugins/fastify-plugin.mjs
```diff
@@ -0,0 +1,13 @@
...
+
+function pluginMjs (f, opts, next) {
+  f.get('/mjs-fastify-plugin', (request, reply) => {
+    reply.send({ hello: 'world' })
+  })
+
+  next()
+}
+
+export default fastifyPlugin(pluginMjs)
```
mcollina: I'm not sure this is actually valid with experimental modules, while I think it is valid with `esm`.

mcollina:
Ping! Any progress here? ESM is now unflagged in Node 13, and it would be great to add support for this.

jhechtf:
Checking for an update on this -- last update was December 6th (~2 months)?

SerayaEryn:
I am working on it. I have a problem with modules which are declared with `"type":"module" in the package.json.

jhechtf:
I got something running in a `"type":"module"` repo that loads a single layer. My solution required me to use an external module `file-url` in order to convert filenames to a `file://` path (which my Node was complaining about). I'd be more than willing to try and help, I just don't know where exactly you're hitting the wall.

SerayaEryn:
I am getting the following error when trying to import the `.js` files from `./test/mjs/packagejson` using `import()`. I should be able to import them since there are `package.json` files with `{ "type": "module" }` present.
It is weird that the error tells me that I used `require()` what I didn't:
```
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /Users/denisfaecke/Documents/fastify-autoload/test/mjs/packagejson/plugin.js
require() of ES modules is not supported.
require() of /Users/denisfaecke/Documents/fastify-autoload/test/mjs/packagejson/plugin.js from /Users/denisfaecke/Documents/fastify-autoload/index.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename plugin.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /Users/denisfaecke/Documents/fastify-autoload/test/mjs/packagejson/package.json.

    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1167:13)
```

jhechtf:
Interesting. I'll see if I can look into it later and come up with any
fixes.

jhechtf:
I got through that issue but have run into another issue that I think might be a genuine bug with the implementation of ESM in Node. I'm going to see if i can recreate the issue without using fastify just to make sure I'm not getting wrapped around in other issues.

jhechtf:
Yet on my laptop I have the same errors as before. Weird.

What I found to fix it on my laptop was simply removing the `package.json` from `test/mjs/packagejson/subfolder/` Not really clear on why that file was even there, but without it the tests run just fine.

jhechtf:
In the interim until support gets added into the main module I've created a rough clone that is specifically for ESM projects called 'fastify-esm-autoload'. The rewrites in this PR were quite extensive and it made it a bit harder for me to look through them. What with the Coronavirus keeping us home I'm hoping to get some extra time to look through and see if I can help with the issues being addressed here.

mcollina:
Superseded by https://github.com/fastify/fastify-autoload/pull/76

PR #57, opened 2019-12-06
Automatically wrap full routes into fastify plugins

### LINKED ISSUE #54: Automatically wrap route files
## 🚀 Feature Proposal

We're migrating an API to fastify and somehow the boilerplate

```
module.exports = function (fastify, opts, next) {
  fastify.route({
    method: 'GET',
    url: '/',
    handler
  })
}
```

feels a bit clunky. I'd like to propose just the route content in a file:

```
module.exports = {
  method: 'GET',
  url: '/',
  handler
}
```

For most cases this is totally sufficient. If additional config/options are required, a regular plugin/route function can be exported.

## Motivation

As mentioned above, the plugin structure feels a bit clunky in each and every file.

## Example

The default type is a function and the new proposal would be an object (method is always required anyway), so I'm thinking somewhere along the lines of (https://github.com/fastify/fastify-autoload/blob/master/index.js#L110):

```
const content = require(file)
let plugin

if (content && typeof content === 'object' && content.method) {
  plugin = function (fastify, opts, next) {
    fastify.route(content)
    next()
  }
} else {
  plugin = content
}
```

This is non-breaking, but to be safe it could also be flagged with an new option "wrapRoutes" or something like that.

I'm testing this locally, but I'm currently getting an error when running `npm test`?

```
test/error.js ......................................... 1/2
  not ok should match pattern provided
    found: >-
      Unexpected token '}' at
      /Users/patrick/Sites/fastify-autoload/test/error/lib/a.js:6
    pattern: '/Unexpected token \} at .*\/test\/error\/lib\/a.js:6/'
    at:
      line: 14
      column: 5
      file: test/error.js
    stack: |
      test/error.js:14:5
      Object._encapsulateThreeParam (node_modules/avvio/boot.js:408:13)
      Boot.callWithCbOrNextTick (node_modules/avvio/boot.js:339:5)
      release (node_modules/fastq/queue.js:127:16)
      Object.resume (node_modules/fastq/queue.js:61:7)
      node_modules/avvio/boot.js:155:20
      node_modules/avvio/plugin.js:178:7
      done (node_modules/avvio/plugin.js:136:5)
      check (node_modules/avvio/plugin.js:147:7)
    source: |
      t.match(err.message, /Unexpected token \} at .*\/test\/error\/lib\/a.js:6/)
```

PatrickHeneise:
Closes #54 

The idea is to reduce the routing files to just the route schema, without wrapping everything into a fastify plugin in every file. It's a cosmetic change.

I'm missing the docs part. Let me know if everything is good to go, then I'll add the part to the README.

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -107,7 +107,18 @@ module.exports = function (fastify, opts, next) {
         }
 
         try {
-          const plugin = require(file)
+          const content = require(file)
+          let plugin
+
+          if (content && typeof content === 'object' && content.method) {
+            plugin = function (fastify, opts, next) {
+              fastify.route(content)
+              next()
+            }
+          } else {
+            plugin = content
+          }
+
           const pluginOptions = Object.assign({}, defaultPluginOptions)
           const pluginMeta = plugin[Symbol.for('plugin-meta')] || {}
           const pluginName = pluginMeta.name || file
--- a/index.js
+++ b/index.js
@@ -109,8 +109,9 @@ module.exports = function (fastify, opts, next) {
         try {
           const content = require(file)
           let plugin
-
-          if (content && typeof content === 'object' && content.method) {
+          if (content && typeof content === 'object' &&
+          Object.prototype.toString.apply(content) === '[object Object]' &&
+          Object.prototype.hasOwnProperty.call(content, 'method')) {
             plugin = function (fastify, opts, next) {
               fastify.route(content)
               next()
--- a/index.js
+++ b/index.js
@@ -109,7 +109,7 @@ module.exports = function (fastify, opts, next) {
         try {
           const content = require(file)
           let plugin
-          if (content && typeof content === 'object' &&
+          if (content &&
           Object.prototype.toString.apply(content) === '[object Object]' &&
           Object.prototype.hasOwnProperty.call(content, 'method')) {
             plugin = function (fastify, opts, next) {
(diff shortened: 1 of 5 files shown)


PatrickHeneise:
Absolutely not. Wanted to get some feedback and wrap up the code & tests first. Docs are coming in a moment.

index.js
```diff
@@ -107,7 +107,18 @@ module.exports = function (fastify, opts, next) {
         }
 
         try {
-          const plugin = require(file)
+          const content = require(file)
+          let plugin
+
+          if (content && typeof content === 'object' && content.method) {
```
jsumners: This check for object-ness is not suffcient. An array will pass this check. You should `Object.prototoype.toString.apply(content) === true`.

PatrickHeneise: What do you think of:

```
Object.prototype.toString.apply(content) === '[object Object]' &&
Object.prototype.hasOwnProperty.call(content, 'method')
```

?

jsumners: Yes, that's what I meant 🤦‍♂. It's still early morning for me over here.

PatrickHeneise: No worries :) thanks for the hint. I thought about that before but then though this might be already sufficient but you're right. Adding both checks now.

index.js
```diff
@@ -107,7 +107,19 @@ module.exports = function (fastify, opts, next) {
         }
 
         try {
-          const plugin = require(file)
+          const content = require(file)
+          let plugin
+          if (content && typeof content === 'object' &&
```
jsumners:
```suggestion
          if (content &&
```

PR #60, opened 2020-01-04
support plugin.autoConfig

davidmarkclements:
This allows plugins to declare their own default options, keys are overridden by custom options provided to autoload. Using the `options` namespace corresponds with `fastify-cli` `-o` flag functionality for the entry point plugin.

Beyond the declarative and convenience benefits this also supports a novel pattern where ecosystem plugins are configured in their own files - which for large scale applications may be easier to manage than in a single entry point. 

For instance consider the following: 

```js
const helmet = require('fastify-helmet')
helmet.options = { referrerPolicy: true }
module.exports = helmet
```

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -120,7 +120,7 @@ module.exports = function (fastify, opts, next) {
             plugin = content
           }
 
-          const pluginOptions = Object.assign({}, defaultPluginOptions)
+          const pluginOptions = Object.assign({}, plugin.options || {}, defaultPluginOptions)
           const pluginMeta = plugin[Symbol.for('plugin-meta')] || {}
           const pluginName = pluginMeta.name || file
 
--- a/index.js
+++ b/index.js
@@ -3,7 +3,7 @@
 const fs = require('fs')
 const path = require('path')
 const steed = require('steed')
-
+const kBaseOptions = Symbol.for('base-options')
 module.exports = function (fastify, opts, next) {
   const defaultPluginOptions = opts.options
 
@@ -119,10 +119,14 @@ module.exports = function (fastify, opts, next) {
           } else {
             plugin = content
           }
-
-          const pluginOptions = Object.assign({}, plugin.options || {}, defaultPluginOptions)
+          const pluginConfig = plugin.options || plugin[kBaseOptions] || {}
+          const pluginOptions = Object.assign({}, pluginConfig, defaultPluginOptions)
           const pluginMeta = plugin[Symbol.for('plugin-meta')] || {}
           const pluginName = pluginMeta.name || file
+
+          if (typeof plugin.options === 'object') {
+            plugin.options = plugin[kBaseOptions]
+          }
 
           if (opts && !plugin.autoPrefix) {
             plugin.autoPrefix = opts.prefix
--- a/index.js
+++ b/index.js
@@ -3,7 +3,7 @@
 const fs = require('fs')
 const path = require('path')
 const steed = require('steed')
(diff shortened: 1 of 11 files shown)


mcollina:
I think the example above mutates the original plugin, and as a result if we did that in two places one would override the other. Can you change it so that does not happen?

davidmarkclements:
@mcollina ok I've found an elegant way to handle this, it depends on this PR in fastify-plugin https://github.com/fastify/fastify-plugin/pull/82

There's two things we're trying to achieve here

* a plugin declaring base options on itself that fastify autoload will honour 
* the ability to configure required plugins via fastify autoload

This is the approach:

* fastify plugin checks for a `plugin.options` key and references it with a `plugin[Symbol.for('base-options')]` key
* fastify autoload will merge the `plugin.options` into the options 
* fastify autoload will then reset `plugin.options` back to `plugin[Symbol.for('base-options')]` - or otherwise `undefined` to clear it

This allows for a single usage configuration options property while also supporting any plugin module itself having a default fallback options key

package.json
```diff
@@ -37,7 +37,7 @@
     "@typescript-eslint/eslint-plugin": "^2.0.0",
     "@typescript-eslint/parser": "^1.12.0",
     "fastify": "^2.10.0",
-    "fastify-plugin": "^1.5.0",
+    "fastify-plugin": "https://github.com/davidmarkclements/fastify-plugin",
```
davidmarkclements: todo: update when/if https://github.com/fastify/fastify-plugin/pull/82 lands

index.js
```diff
@@ -3,7 +3,7 @@
 const fs = require('fs')
 const path = require('path')
 const steed = require('steed')
-
+const kBaseOptions = Symbol.for('base-options')
```
jsumners: Can this be namespaced? Maybe `Symbol.for('fastify-plugin-default-options')` or something?

davidmarkclements:
> How is this different than?

in that example baseOptions is locally scoped and therefore private, and baseOptions will always apply. 

In this case the options will only be mixed in when using autoload, and more importantly can be set externally (on the plugin instance).

You could actually also do both, you could have your baseOptions example *and* use plugin.options, so when autoloaded it would be functionally equivalent to `Object.assign({}, baseOptions, Object.assign({}, plugin.options, autoloadOptsOptionsKey))`

The more we talk about this more I think it should actually be called `autoConfig` to make things clearer, base options is, I think, perhaps an inappropriate name for what this is.

In terms of utility this allows for shorthand object-based configuration, and it's comparable with autoload support for exporting route declarations. It's essentially boilerplate reduction.

In terms of ergonomics the boilerplate reduction fits in a wider architectural strategy that involves standardising and distributing configs as atomic pieces across many projects.

mcollina:
I think calling this autoConfig is actually better.

mcollina:
or autoOptions (but not autobots)

jsumners:
image

davidmarkclements:
ok. so @mcollina I've renamed options to autoConfig, agreed this feels much better.

Also I thought this through some more and concluded there's no benefit to a "base options" (or "base config") object - since using autoConfig is a one-time operation, and plugin libraries should handle their default options their own way (as @jsumners hinted at).

Closing https://github.com/fastify/fastify-plugin/pull/82 because of this, thanks everyone for sharpening this with excellent feedback

package.json
delvedor: Can you restore this one?

davidmarkclements:
@delvedor thank you - done!

PR #64, opened 2020-02-17
github actions

SerayaEryn:
#### Checklist


.github/workflows/ci.yml
```diff
@@ -0,0 +1,26 @@
...
+
+      - name: Use Node.js
+        uses: actions/setup-node@v1
+        with:
+          node-version: ${{ matrix.node-version }}
+
+      - name: Install
+        run: |
+          npm install
+      - name: Run Tests
```
RafaelGSS: Maybe, can be good add the lint check

Eomm: the `test` run has the `standard` check already :+1:

RafaelGSS:
I think that we don't need to support Node 6.

PR #66, opened 2020-02-19
Ignore TypeScript definition files

### LINKED ISSUE #65: Type definition files should not attempt to be loaded
## 🐛 Bug Report

When using `includeTypeScript: true`, fastify-autload will attempt to load type definition files (`.d.ts`), causing syntax errors, such as _Cannot use import statement outside a module_.

This happens when Node is started into a directory built from TypeScript sources into a set of `.js` & `.d.ts` files.

## To Reproduce

Reproduction repository available here:
https://github.com/franky47/fastify-autoload-65

## Expected behavior

Since Node.js cannot understand type definition files, and they contain no runtime code, I believe the extension `.d.ts` should be ignored, even when `includeTypeScript` is true.

## Your Environment

- *node version*: 13.7.0
- *fastify version*: 2.11.0
- *fastify-autoload version*: 1.2.0
- *os*: macOS

franky47:
Closes #65.

#### Checklist

- [x] run `npm run test` and `npm run benchmark` _(no benchmark in this plugin)_

### DIFF
--- a/index.js
+++ b/index.js
@@ -12,7 +12,7 @@ module.exports = function (fastify, opts, next) {
     ? /^index\.(ts|js)$/im
     : /^index\.js$/im
   const scriptPattern = opts.includeTypeScript
-    ? /\.(ts|js)$/im
+    ? /((?<!\.d)\.ts|\.js)$/im // For .ts files, ignore .d.ts, see #65
     : /\.js$/im
 
   function enrichError (err) {
--- a/index.js
+++ b/index.js
@@ -12,7 +12,7 @@ module.exports = function (fastify, opts, next) {
     ? /^index\.(ts|js)$/im
     : /^index\.js$/im
   const scriptPattern = opts.includeTypeScript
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -12,7 +12,7 @@ module.exports = function (fastify, opts, next) {
     ? /^index\.(ts|js)$/im
     : /^index\.js$/im
   const scriptPattern = opts.includeTypeScript
-    ? /\.(ts|js)$/im
+    ? /((?<!\.d)\.ts|\.js)$/im // For .ts files, ignore .d.ts, see #65
```
Eomm: CI is failing for

> not ok Invalid regular expression: /((?<!\.d)\.ts|\.js)$/: Invalid group

Could you change the regexp to avoid the Negative lookbehind?

franky47: Yeah I did that on my fork, but the commit did not make it here, that's weird..

mcollina:
Would you mind to add a unit test?

franky47:
I added a .d.ts file that crashed the tests before updating the regex, do you have another test in mind ?

PR #72, opened 2020-04-30
51 load subfolders

alemagio:
#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -26,6 +26,68 @@ module.exports = function (fastify, opts, next) {
     return err
   }
 
+  function loadDirs (dirs, cb, plugins, prefix) {
+    if (dirs.length === 0) {
+      cb(null, plugins)
+      return
+    }
+
+    const toLoad = dirs.pop()
+    fs.stat(toLoad, (err, stat) => {
+      if (err) {
+        cb(err)
+        return
+      }
+
+      if (stat.isDirectory()) {
+        fs.readdir(toLoad, (err, files) => {
+          if (err) {
+            cb(err)
+            return
+          }
+
+          prefix = path.join(prefix, toLoad.split(path.sep).pop())
+          const fileList = files.join('\n')
+          // if the directory does not contain a package.json or an index,
+          // load each script file as an independend plugin
+          if (
+            !packagePattern.test(fileList) &&
+            !indexPattern.test(fileList) &&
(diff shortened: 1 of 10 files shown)


alemagio:
As I said in the issue, still working on it but I think it could be useful to collect some feedback in the meantime.

index.js
```diff
@@ -26,6 +26,68 @@ module.exports = function (fastify, opts, next) {
     return err
   }
 
+  function loadDirs (dirs, cb, plugins, prefix) {
```
Eomm, line 29: I would put the `cb` at the end

index.js
```diff
@@ -26,6 +26,68 @@ module.exports = function (fastify, opts, next) {
...
+              }
+            }
+          } else {
+            plugins.push({
+              // skip directories without script files inside
+              skip: files.every(name => !scriptPattern.test(name)),
+              file: toLoad
+            })
+          }
+          loadDirs(dirs, cb, plugins, prefix)
```
Eomm, line 78: Just to share: I'm asking myself if this should be an opt-in option or just "load all the folders".

Right now I think it is fine 👍

alemagio: reading the issue comments my idea was:
add an option called `recursive` with boolean value, if `true` it loads recursively all dirs

to customize the folder loading the user can add an `autoload.json` file with some options.

Is that correct?

Eomm: I like the 'recursive' flag

I would avoid the autoload.json config.
I think, for example an option sync function filter(path) {} where the user return true or false if he wants to load that folder for example

But I would go through steps 👍

mcdonnelldean:
Can we add a test to ensure this works with the auto-routing functionality already available too.

alemagio:
Added a small fix since I noticed that it did not apply the right prefix in case of multiple subfolders.
Added a test for that scenario and added a test for autoroute @mcdonnelldean , please let me know if it is what you meant.

alemagio:
Is there something i did wrong that broke CI?

mcollina:
I'm not sure!

I'm actually close to land https://github.com/fastify/fastify-autoload/pull/76 which is a significant refactor to simplify esm and ts support. You might want to target that.

alemagio:
Sure

mcollina:
This was implemented as part of #76.

PR #74, opened 2020-05-12
update: add support esm

RafaelGSS:
Hello!

I don't found in any doc the support at ESM on `fastify-autoload`. So I created it.
However, I don't have an idea of how to test it and make the support in previous versions. Any idea?

#### Checklist

- [X] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)

### DIFF
--- a/fastify-autoload.d.ts
+++ b/fastify-autoload.d.ts
@@ -7,7 +7,7 @@ export interface AutoloadPluginOptions {
   dir: string
   ignorePattern?: RegExp | string
   includeTypeScript?: boolean
-  esmModules?: boolean
+  loadESM?: boolean
   options?: FastifyPluginOptions
 }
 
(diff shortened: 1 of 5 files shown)


fastify-autoload.d.ts
```diff
@@ -7,6 +7,7 @@ export interface AutoloadPluginOptions {
   dir: string
   ignorePattern?: RegExp | string
   includeTypeScript?: boolean
+  esmModules?: boolean
```
jsumners, line 10: This is redundant. “ESM” === “ECMA Script Modules”. So this code is saying “ECMA Script Modules Modules”.

RafaelGSS: `esModules` sounds good for you?

jsumners: `loadESM: true`?

RafaelGSS:
This PR is overridden by #76 ?

jsumners:
I think so? I don't know. Please coordinate with the author in that thread.

PR #76, opened 2020-05-13
ESM support and async/await refactor

thomheymann:
This PR adds support for ES modules, includes a refactor to use async/await instead of callback based directory scanning and increases test coverage for different syntax styles (commonjs, module, typescript):

- Automatically loads in `.cjs` files as CommonJS, `.mjs` files as ES modules and `.ts` files as TypeScript
- Automatically detects whether `.js` files should be loaded in as CommonJS or ES modules according to `type` field of parent `package.json` (this is inline with ES module spec)
- ⚠️ Dropped support for loading in folders that contain a `package.json` file as a plugin. Unfortunately this is not compatible with ES modules - Instead create an `index.js` file to treat the entire folder as a single plugin
- ⚠️ Dropped support for `includeTypescript` option. Typescript files are now loaded in automatically if present. If for any reason there are `.ts` files in a folder but they shouldn't be loaded in then they can be excluded using the existing `ignorePattern` option
- Will throw an error if `.ts` files are present but process isn't running `ts-node`
- Will throw an error if scripts use `module` syntax but node version doesn't support modules
- Will throw an error if `package.json` file is found inside the plugin directory
- Tests have been refactored to
- ⚠️ Gotcha: JavaScript files written using Babel "ES module" syntax are NOT valid ES modules and can't be loaded by Node.js. Either update those files according to the spec or continue to transpile them using Babel before loading them in with `fastify-autoload`.
- ⚠️ Gotcha: TypeScript can use ES module syntax but uses CommonJS behaviour (same as Babel). Those will continue to be loaded using `require` rather than `import`

#### Checklist


jsumners:
How does this compare with #74 ?

test/module/basic.js
```diff
@@ -0,0 +1,162 @@
+'use strict'
```
mcollina, line 1: There is no need for this in esm.

test/module/basic/app.js
```diff
@@ -0,0 +1,32 @@
+'use strict'
```
mcollina, line 1: There is no need for 'use strict' in esm

index.js
```diff
@@ -5,8 +5,10 @@ const fs = require('fs')
 const path = require('path')
 const readDirectory = promisify(fs.readdir)
 const pkgUp = require('pkg-up')
+const semver = require('semver')
 
 const typescriptSupport = Symbol.for('ts-node.register.instance') in process
+const moduleSupport = semver.satisfies(process.version, '>=14')
```
jsumners, line 11:
```suggestion
const moduleSupport = semver.satisfies(semver.coerce(process.version), '>=14')
```

Should it be that instead? I believe that will allow for Node 15, 16, .... N.

thomheymann:
@jsumners Sorry, github was acting up and now i can't find your suggestion. I think nodejs uses semver so `process.version` will always return a valid semver string but might be worth adding for prerelease versions. I'll make the suggested change locally instead

jsumners:
> @jsumners Sorry, github was acting up and now i can't find your suggestion. I think nodejs uses semver so `process.version` will always return a valid semver string but might be worth adding for prerelease versions. I'll make the suggested change locally instead

I deleted it. I realized after posting that the coerce is only needed for `-rc.x` style releases and that Node core doesn't do that.

mcollina:
It needs more work on the README.

thomheymann:
Documentation has been added

README.md
```diff
@@ -1,231 +1,204 @@
...
 
-Require all plugins in a directory.
+Convenience plugin for Fastify that loads all plugins found in a directory and automatically configures routes matching the folder structure.
 
-## Install
+## Installation
 
 ```
-npm i fastify fastify-autoload
+npm i --save fastify fastify-autoload
```
davidmarkclements, line 8: save it automatic, we don't need to specify

README.md
```diff
@@ -1,231 +1,204 @@
...
 
-```js
-'use strict'
+Fastify server that automatically loads in all plugins from the `plugins` directory:
 
-const Fastify = require('fastify')
-const AutoLoad = require('fastify-autoload')
+```js
+const fastify = require('fastify')
+const fastifyAutoload = require('fastify-autoload')
```
davidmarkclements, line 17:
```suggestion
const autoload = require('fastify-autoload')
```

README.md
```diff
@@ -1,231 +1,204 @@
 # fastify-autoload
```
davidmarkclements, line 1: I'm not seeing any esm examples in the readme (sorry if I missed them) - can we please add esm examples (especially for autoPrefix and autoConfig)

index.js
```diff
@@ -1,207 +1,185 @@
...
-          next(enrichError(err))
-          return
-        }
+      if (type === 'module' && !moduleSupport) {
+        throw new Error(`fastify-autoload cannot import plugin at '${file}'. Your version of node does not support ES modules. To fix this error upgrade to Node 14 or use CommonJS syntax.`)
       }
+      accumulator.push({ file, type, prefix })
+    }
+  }
+  await Promise.all(directoryPromises)
```
davidmarkclements, line 97: if we use `Promise.anySettled` here we can collect all failures instead of failing on the first, might be useful

index.js
```diff
@@ -1,207 +1,185 @@
...
+
+  fastify.register(plugin, options)
+  meta.registered = true
+}
+
+function wrapRoutes (content) {
+  if (content &&
+    Object.prototype.toString.call(content) === '[object Object]' &&
+    Object.prototype.hasOwnProperty.call(content, 'method')) {
+    return function (fastify, opts, next) {
```
davidmarkclements, line 167: why not use async/await here as well? wouldn't  a throw from fastify.route then propagate up?

package.json
```diff
@@ -3,13 +3,15 @@
...
-  "typings": "fastify-autoload.d.ts",
+  "types": "fastify-autoload.d.ts",
   "scripts": {
-    "lint": "standard | snazzy && npm run lint:typescript",
-    "lint:fix": "standard --fix",
-    "lint:typescript": "standard --parser @typescript-eslint/parser --plugin @typescript-eslint/eslint-plugin ./*.ts",
-    "test": "standard | snazzy && tap test/*.js && npm run typescript",
-    "typescript": "tsc --project ./test/types/tsconfig.json && tap test/*.ts"
+    "lint": "standard | snazzy",
+    "with": "",
```
davidmarkclements, line 9: empty script field?

test/commonjs/basic/app.js
```diff
@@ -2,7 +2,7 @@
 
 const fs = require('fs')
 const path = require('path')
-const AutoLoad = require('../..')
+const AutoLoad = require('../../../')
```
davidmarkclements, line 5:
```suggestion
const autoload = require('../../../')
```

test/commonjs/error.js
```diff
@@ -0,0 +1,33 @@
+'use strict'
+
+const t = require('tap')
+const Fastify = require('fastify')
```
davidmarkclements, line 4: (I know these we're capitalised before but it violates the rule of constructor/class pascal case)

```suggestion
const fastify = require('fastify')
```

test/module/basic/defaultPrefix/prefixed.js
```diff
@@ -0,0 +1,11 @@
...
+
+export default function (f, opts, next) {
+  f.get('/', (request, reply) => {
+    reply.send({ prefixed: true })
+  })
+
+  next()
+}
+
+export const autoPrefix = '/prefixed'
```
davidmarkclements, line 11: this is great, I like this a lot as an interface

README.md
davidmarkclements: actually I wonder if it might be word having a `README.esm.md` which is an esm version of this readme and then link to it from here

eventually we'd make README.esm.md the README.md and move README.md to README.cjs.md

index.js
mcollina: We can propagate only one error up anyway. What should we do with those errors?

mcollina:
@davidmarkclements done, lmk what you think

index.js
```diff
@@ -1,207 +1,184 @@
...
-    ? /((^.?|\.[^d]|[^.]d|[^.][^d])\.ts|\.js)$/im // For .ts files, ignore .d.ts
-    : /\.js$/im
-
-  function enrichError (err) {
-    // Hack SyntaxError message so that we provide
-    // the line number to the user, otherwise they
-    // will be left in the cold.
-    if (err instanceof SyntaxError) {
-      err.message += ' at ' + err.stack.split('\n')[0]
+const readDirectory = promisify(fs.readdir)
```
Eomm, line 6: We could use `require('fs').promises` instead of promisify just to be stricter (it is supported from node 10+)

mcollina:
@Eomm done.

Issue #78, opened 2020-06-03
Import all plugins in parallel


mcollina:
In https://github.com/fastify/fastify-autoload/blob/41983a1ca5f06c7c73affc718e82eeaeacf62b80/index.js#L21-L30, the plugins are loaded one at a time.

Given `import()` is async, it's possible to parallelize it so they would load faster.

iainjreid:
Submitted a PR to apply this change! 👍

mcollina:
Good work!

PR #80, opened 2020-06-05
Import plugins in parallel to reduce load times

iainjreid:
This PR parallelises the loading of plugins, as per #78.


mcollina:
Thanks!

PR #92, opened 2020-07-15
Doesn't load the modules next to an index.js file (fix #90)

### LINKED ISSUE #90: Need some clarification on the way direcotries with index.js are handled
From the doc:

> Each script file within a directory is treated as a plugin unless the directory contains an index file (e.g. index.js). In that case only the index file will be loaded.

In practice this is true only when the directory containing the index.js doesn't include other sub-directories. [In this case](https://github.com/fastify/fastify-autoload/blob/c1d31a7d72b37814f6c1214ee968625ed9d835e9/index.js#L66) the others plugins in the parent-directory and in the sub-directories are also loaded.

I don't know if its a bug or if the documentation isn't up-to-date. IMO adding a sub-directory shouldn't change the way the others plugins in the parent directory are handled.

Thanks!

rackaam:
#### Checklist

--------------------------------------------
This PR assume that the desired behaviour for a directory containing an `index.js` file is to not load the files next to the `index.js` but load the modules in the potential sub-directories next to `index.js`:
```
routes
├── index.js // loaded
├── routeA.js // ignored
└── sub
    └── routeB.js // loaded
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ Autoload can be customised using the following options:
 
 - `dir` (required) - Base directory containing plugins to be loaded
 
-  Each script file within a directory is treated as a plugin unless the directory contains an index file (e.g. `index.js`). In that case only the index file will be loaded.
+  Each script file within a directory is treated as a plugin unless the directory contains an index file (e.g. `index.js`). In that case only the index file (and the potential sub-directories) will be loaded.
   
   The following script types are supported:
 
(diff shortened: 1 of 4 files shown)


mcollina:
would you mind to fix linting?

```

/home/runner/work/fastify-autoload/fastify-autoload/test/commonjs/basic.js
  194:1  error  Trailing spaces not allowed
```

rackaam:
Fixed but npm is down so the tests are failing

mcollina:
Thanks!

PR #93, opened 2020-07-15
add snippet to allow loading multiple directories with one call

ferco0:
Like this:
```js
...
app.register(autoLoad, [
  { dir: join(__dirname, 'plugins')},
  { dir: join(__dirname, 'routes')},
])
....
```
#### Checklist

- [x] tests and/or benchmarks are included // I think this don't need a test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,17 @@ Folder structure:
 └── app.js
 ```
 
+**You can pass the options like as an array to load multiple instances.**
+
+```js
+...
+app.register(autoLoad, [
+  { dir: join(__dirname, 'plugins')},
+  { dir: join(__dirname, 'routes')},
+])
+....
+```
+
 ## Global Configuration
 
 Autoload can be customised using the following options:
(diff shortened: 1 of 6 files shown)


ferco0:
Thanks! would it be good to have the both ways?

mcollina:
I don't really understand the need.. if you need to pass different options, why do it when you can issue two register calls?

ferco0:
For example: If, for any reason, we need a different ignore pattern for each directory or different prefix.

mcollina:
Why is doing multiple `app.register(autoload)` not possible then?

ferco0:
does, but why not by an array? is ready and anyone could prefer it like that, I would prefer. I like to avoid method calls as much as I can in the application code.

mcollina:
I do not think this is needed.

ferco0:
could be a performance increase, avoiding fastify.register call?

mcollina:
not really, that's pretty cheap

Issue #82, opened 2020-06-26
pluginMetadata is empty in v1.2.2


JacopoDaeli:
https://github.com/fastify/fastify-autoload/blob/67d78db66c4ab7b6d593e9dbbecc2b22d49975ed/index.js#L124

Hello, dependencies loading are not respected because it seems that `plugin[Symbol.for('plugin-meta')]` is always `undefined`.

To create my plugins I'm using `fastify-plugin@1.6.1`.

For instance when I print `plugin` in some of my code I get:
```js
{
  default: [Function: connectDynamoDB] {
    [Symbol(skip-override)]: true,
    [Symbol(fastify.display-name)]: 'dynamodb-connector',
    [Symbol(plugin-meta)]: {
      name: 'dynamodb-connector',
      fastify: '2.x',
      decorators: [Object],
      dependencies: [Array]
    }
  }
}
```

Therefore I believe we should change to `const pluginMeta = plugin.default[Symbol.for('plugin-meta')] || {}`.

Does it make sense to fix this and publish a `v1.2.3`?
Although, I'm not sure it is still possible (cc. @mcollina).

mcollina:
Are you transpiling with babel or typescript (or any other lib)?
Please send a PR! Can you also check if this happens with master? It has a different codebase, so we cannot backport easily.

JacopoDaeli:
@mcollina the issue is not in `master`. It looks like `v3.x` has not the problem.
However folks using `fastify@2.x` which need to use `fastify-autoload@1.x` (otherwise we would get some `Typescript` errors. We could create a separate branch for autoload `1.x`, starting from [this commit](https://github.com/fastify/fastify-autoload/commit/67d78db66c4ab7b6d593e9dbbecc2b22d49975ed), and push there the fix?

JacopoDaeli:
Oh forgot to mention. The issue happens both when transpiring using `tsc`, but also when using `node-ts` directly.

mcollina:
I think it's due to some of the options you have configure ts with.

Feel free to send a PR, I've created the branch: https://github.com/fastify/fastify-autoload/tree/v1.x

PR #95, opened 2020-07-25
add support to detect ts-jest environment

### LINKED ISSUE #86: Jest -> ts-jest + Typescript does not appear to be working
## 🐛 Bug Report

In Fastify-Autoload 1.x jest tests + typescript worked with fastify-autoload with seemingly no issue as long as the `includeTypescript: true` configuration was passed.  With the new version it appears that has been removed in favor of automatically supporting ts-node.  I don't 100% know everything about ts-jest but I do not believe it uses ts-node and yet it worked before.  It seems like this should still work.

## To Reproduce

clone this repo - https://github.com/wolfejw86/fastify-typescript-starter

run `npm install`
run `npm run test`

see that fasitify-autoload is working correctly in https://github.com/wolfejw86/fastify-typescript-starter/blob/master/src/app.ts

In the same repo - upgrade fastify-swagger, fastify, fastify-plugin, and fastify-autoload all to the latest version by switching to the branch in progress(the 3.0 migration for fastify + typescript).
`git checkout fastify-upgrade`
`npm install`
`npm run test`

Notice fastify-autoload will no longer load any plugins.

```
fastify-autoload cannot import plugin at '/Users/johnwolfe/Desktop/repositories/fastify-typescript-starter/src/plugins/globalErrorHandler.ts'. To fix this error compile TypeScript to JavaScript or use 'ts-node' to run your app.

      at findPlugins (node_modules/fastify-autoload/index.js:87:15)
      at fastifyAutoload (node_modules/fastify-autoload/index.js:19:19)
```

Edit:
Additionally if I hardcode this line:
image

~It _mostly_ works, but doesn't load an `index.ts` file in the services folder from:~ EDIT: it works fine - I did not correctly understand the new behavior of how autoload plugin handles nested folders
~I am waiting for `fastify.ready()` but it seems like it's firing before the second call to the autoload plugin has finished resolving or something.~ EDIT: It loads all modules correctly according to how Autoload uses nested folders as pathnames now.

Any help is appreciated :)

ferco0:
#### Checklist

Fixes #86 

- [x] run `npm run test` ~~and `npm run benchmark`~~
- [x] tests ~~and/or benchmarks are~~ included

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "unit:without-modules": "tap test/commonjs/*.js test/*/*.ts",
     "test": "npm run lint && npm run unit && npm run typescript && npm run typescript:jest",
     "typescript": "tsd",
-    "typescript:jest": "npx jest"
+    "typescript:jest": "jest"
   },
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -9,8 +9,9 @@
     "unit": "if semver `node --version` --range '>=14 || >= 12.17.0 < 13.0.0'; then npm run unit:with-modules; else npm run unit:without-modules; fi",
     "unit:with-modules": "tap test/*/*.js test/*/*.ts",
     "unit:without-modules": "tap test/commonjs/*.js test/*/*.ts",
-    "test": "npm run lint && npm run unit && npm run typescript",
-    "typescript": "tsd"
+    "test": "npm run lint && npm run unit && npm run typescript && npm run typescript:jest",
+    "typescript": "tsd",
+    "typescript:jest": "npx jest"
   },
   "repository": {
     "type": "git",
@@ -34,14 +35,17 @@
   },
   "homepage": "https://github.com/fastify/fastify-autoload#readme",
   "devDependencies": {
+    "@types/jest": "^26.0.7",
     "@types/node": "^14.0.1",
     "fastify": "^3.0.0",
     "fastify-plugin": "^2.0.0",
     "fastify-url-data": "^3.0.1",
+    "jest": "^26.1.0",
     "pre-commit": "^1.2.2",
     "snazzy": "^8.0.0",
     "standard": "^14.3.3",
     "tap": "^14.0.0",
+    "ts-jest": "^26.1.3",
     "tsd": "^0.13.1"
   },
   "dependencies": {
@@ -56,5 +60,16 @@
   },
   "tsd": {
     "directory": "test/typescript/definitions"
+  },
+  "jest": {
+    "preset": "ts-jest",
+    "testEnvironment": "node",
+    "rootDir": "./test/typescript-jest",
+    "testMatch": [
+      "**/*.test.ts"
+    ],
+    "transform": {
+      "^.+\\.(ts|tsx)$": "ts-jest"
+    }
   }
 }
--- a/test/typescript-jest/basic/app.ts
+++ b/test/typescript-jest/basic/app.ts
@@ -1,0 +1,13 @@
+import { FastifyPlugin } from 'fastify'
+import { join } from 'path'
+const fastifyAutoLoad = require('../../../')
+
+const app: FastifyPlugin = function (fastify, opts, next): void {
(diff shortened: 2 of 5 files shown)


mcollina:
This is great! Can you add a link/reference where you
 did find this?

ferco0:
Thanks! I dumped the process object and looked for something that could help.

mcollina:
could you add a simple test with jest as well? I don't want this to break without notice on a new release.

package.json
```diff
@@ -9,8 +9,9 @@
     "unit": "if semver `node --version` --range '>=14 || >= 12.17.0 < 13.0.0'; then npm run unit:with-modules; else npm run unit:without-modules; fi",
     "unit:with-modules": "tap test/*/*.js test/*/*.ts",
     "unit:without-modules": "tap test/commonjs/*.js test/*/*.ts",
-    "test": "npm run lint && npm run unit && npm run typescript",
-    "typescript": "tsd"
+    "test": "npm run lint && npm run unit && npm run typescript && npm run typescript:jest",
+    "typescript": "tsd",
+    "typescript:jest": "npx jest"
```
RafaelGSS, line 14:
```suggestion
    "typescript:jest": "jest"
```
You already installed at devDependencies.

test/typescript-jest/basic/app.ts
```diff
@@ -0,0 +1,13 @@
+import { FastifyPlugin } from 'fastify'
+import { join } from 'path'
+const fastifyAutoLoad = require('../../../')
```
mcollina, line 3: Could you use import instead?

ferco0: I don't think this is needed, will need install babel-jest and add more settings.

wolfejw86:
fwiw the property referenced on process.env always comes up `undefined` for me.  I'm not sure this works with all versions of ts-jest 🤔 

Is it documented anywhere that it is setting this variable on the process?

wolfejw86:
With further inspection - this flag `process.env.npm_package_jest_transform_______ts_tsx__` seems to _only_ be set when your jest config settings live inside your package.json file.  If this change goes in as is, it may only support repos that use jest configuration that way which might be a more narrow scope than was intended?  I know in my circumstances (and I believe is common) I've always used a configuration file separate from `package.json`

mcollina:
@wolfejw86 wdyt, should this land or not?

wolfejw86:
I absolutely don't want to take away from the work @ferco0 did - it's really cool to take a deep dive and see how running things through an `npm` command enhances the `process.env` object.  Unfortunately this solution does not solve the referenced issue when running in a repository via a separate jest config file _nor_ does it work when running without the use of an npm command (common in repos that use `yarn` or repos that deploy via node-slim containers where you do `node server.js` to start the app).
Therefore my 2 cents is that since it doesn't fully solve #82 I would not recommend it to land.

ferco0:
I installed yarn to test it, I put the jest settings in `jest.config.js` and it works for me:



I think that don't need check the Typescript support if executes a script with node command.

wolfejw86:
That’s fair about the node command part. I didn’t realize it worked for
yarn too. The bot working with a heat config file still gets me though

wolfejw86:
Hmmm, I definitely can't reproduce it anymore.  I was probably doing something silly in tandem with testing this out.  It appears to work fine.  With the assumption that you aren't ever going to need to run this with a barebones nodejs command I think it's good to go.  Sorry for the hold up 🤦

ferco0:
@wolfejw86 thanks man, you helped improve the detection! :pray:

johanneslumpe:
@mcollina For what it's worth, this fix does not work for me. `process.env.npm_package_jest_transform_______ts_tsx__` is not defined when I run my tests with a custom config file. Is there a chance we can get the `includeTypeScript` flag back instead?

mcollina:
a PR is always welcomed.

johanneslumpe:
Turns out I cannot reproduce it anymore. Not idea what happened, sorry for the confusion.

Issue #88, opened 2020-07-14
100% code coverage


mcollina:
As titled, only a few lines are missing
```

----------|----------|----------|----------|----------|-------------------|
File      |  % Stmts | % Branch |  % Funcs |  % Lines | Uncovered Line #s |
----------|----------|----------|----------|----------|-------------------|
All files |    97.96 |    95.45 |      100 |    97.89 |                   |
 index.js |    97.96 |    95.45 |      100 |    97.89 |            96,149 |
----------|----------|----------|----------|----------|-------------------|


```

PR #96, opened 2020-07-26
test: add cyclic-dependency and module-error tests

RafaelGSS:
fixes: #88 

#### Checklist

- [X] run `npm run test` and `npm run benchmark`
- [X] tests and/or benchmarks are included
- [X] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)

### DIFF
--- a/test/commonjs/cyclic-dependency/app.js
+++ b/test/commonjs/cyclic-dependency/app.js
@@ -1,0 +1,12 @@
+'use strict'
+
+const path = require('path')
+const AutoLoad = require('../../../')
+
+module.exports = function (fastify, opts, next) {
+  fastify.register(AutoLoad, {
+    dir: path.join(__dirname, 'lib')
+  })
+
+  next()
+}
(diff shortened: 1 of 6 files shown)


mcollina:
can you add a `-100` argument to tap to enforce 100% code coverage?

iandrc:
> can you add a `-100` argument to tap to enforce 100% code coverage?

Just a small clarification. It would be more like `--100` as referred in the [docs](https://node-tap.org/docs/coverage/100/).

RafaelGSS:
> can you add a `-100` argument to tap to enforce 100% code coverage?

Unfortunately, if we add this option the result is non-zero exit code. Because some tests are specifics to `esm`.

For instance when running the unit at v12.13:

image
and this line is [here](https://github.com/RafaelGSS/fastify-autoload/blob/100-coverage/index.js#L110) that are covered to module tests.

Therefore, when we run at v12.17:

image
[here](https://github.com/RafaelGSS/fastify-autoload/blob/100-coverage/index.js#L96) that are coverated as well.

Issue #98, opened 2020-08-12
error with esm on windows


amis-shokoohi:
I was following @mcollina tutorial for Fastify, and I faced with an error related to fastify-autoload plugin:
```sh
Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only file and data URLs are supported by the default ESM loader
    at Loader.defaultResolve [as _resolve] (internal/modules/esm/resolve.js:781:11)
    at Loader.resolve (internal/modules/esm/loader.js:85:40)
    at Loader.getModuleJob (internal/modules/esm/loader.js:229:28)
    at Loader.import (internal/modules/esm/loader.js:164:28)
    at importModuleDynamically (internal/modules/cjs/loader.js:1199:27)
    at exports.importModuleDynamicallyCallback (internal/process/esm_loader.js:30:14)
    at loadPlugin (C:\Users\amis\Desktop\fastify3-test\node_modules\fastify-autoload\index.js:117:5)
    at C:\Users\amis\Desktop\fastify3-test\node_modules\fastify-autoload\index.js:27:12
    at Array.map (<anonymous>)
    at fastifyAutoload (C:\Users\amis\Desktop\fastify3-test\node_modules\fastify-autoload\index.js:26:29) {
  code: 'ERR_UNSUPPORTED_ESM_URL_SCHEME'
}
```
I'm running Node v14.7.0 on Windows 10.
This is how the app looks like:
##### app.js
```javascript
import fastify from 'fastify'
import autoload from 'fastify-autoload'
import { join } from 'desm'

export default function (opts) {
    const app = fastify(opts)

    app.register(autoload, {
        dir: join(import.meta.url, 'routes')
    })

    return app
}
```
##### /routes/hello.js
```javascript
export default async function (app) {
    app.get('/', async function (req, rep) {
        return { hello: 'world' }
    })
}
```

Eomm:
The join from desm is doing this work?

import { dirname } from 'path'; 
import { fileURLToPath } from 'url'; 

const __dirname = dirname(fileURLToPath(import.meta.url)); ... path.join(__dirname, 'public'))) ...

mcollina:
yes, that's the idea

amis-shokoohi:
I've actually tried that but still getting the same error.
I also went through `fastify-autoload/index.js` in debug mode, it seemed to work properly.
The problem is with esm. I rewrote it with cjs, it worked. I was just too excited to use esm.

[...]

PR #99, opened 2020-08-13
Fixed "error with esm on windows #98" issue

amis-shokoohi:
#98 Fix for ESM on Windows


StarpTech:
Enable Windows in the CI to avoid regression.

fox1t:
@amis-shokoohi you can use this matrix to enable testing on all OSes: https://github.com/fastify/fastify/blob/master/.github/workflows/ci.yml#L23

PR #100, opened 2020-08-31
Added support for ts-node-dev

MichielDeMey:
This PR adds support for [ts-node-dev](https://github.com/whitecolor/ts-node-dev) which uses `ts-node` under the hood.
Tested and verified working on my local machine.

From the `ts-node-dev` docs:
> If you need to detect that you are running with ts-node-dev, check if process.env.TS_NODE_DEV is set.

AFAIK, no documentation changes are required as TypeScript is supported out-of-the-box by this module.

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -6,7 +6,7 @@ const { readdir } = require('fs').promises
 const pkgUp = require('pkg-up')
 const semver = require('semver')
 
-const isTsNode = Symbol.for('ts-node.register.instance') in process
+const isTsNode = (Symbol.for('ts-node.register.instance') in process) || !!process.env.TS_NODE_DEV
 const isJestEnviroment = process.env.JEST_WORKER_ID !== undefined
 const hasTsJest = 'npm_package_devDependencies_ts_jest' in process.env
 const typescriptSupport = isTsNode || (isJestEnviroment && hasTsJest)


RafaelGSS:
Could you add a test for it?

MichielDeMey:
Any pointers on how to write tests for this?
Couldn't find a test for the original `ts-node` support in the repo. Might have overlooked it.

mcollina:
ts-node is checked via tap typescript support

MichielDeMey:
> ts-node is checked via tap typescript support

Does this mean that TypeScript is already tested regardless of running through `ts-node`, `ts-node-dev` or others? 🤔

mcollina:
well, definitely not on ts-node-dev, otherwise it would have worked!

MichielDeMey:
Hence this PR. 🙂 

But honestly, I have no clue how to test for this in an automated way.

The only thing preventing `ts-node-dev` from working at this point is the boolean logic at:
https://github.com/fastify/fastify-autoload/blob/master/index.js#L100.
Simply extending the `typescriptSupport` boolean to also check for the `ts-node-dev` environment works perfectly.

Should I go ahead and create a separate issue for writing tests for this?
Maybe the community can help out?

fox1t:
I think this is a needed feature and we should land it as soon as possible. What is not clear to me is why `ts-node.register.instance` is not set in `process`. By the way, we are already missing a test for `ts-node`. 

@MichielDeMey probably testing this would involve running a test file using ts-node and ts-node-dev and checking the variables in `process`/`process.env`.

RafaelGSS:
Sorry for the delay. I agree that's should land as soon and unfortunately, the tests could have a bit of complexity.

However, the `ts-node` should be tested on `npm run typescript:jest`. The problem is that on tests the `ts-node` symbol not exists and `typescriptSupport` is set by:
```js
const typescriptSupport = isTsNode || (isJestEnviroment && hasTsJest /* true */) 
```
And the affirmation of @fox1t: 
> By the way, we are already missing a test for ts-node.

is true.

IMO, we can land this and create an issue to solve both tests

mcollina:
can you folks create such an issue? Thanks.

RafaelGSS:
I created, feel free to update the description if it's not clear.

RafaelGSS (approved):
LGTM.

MichielDeMey:
> What is not clear to me is why `ts-node.register.instance` is not set in `process`.

I was confused as well and did some digging. They only recently added the environment variable option as a solution to checking if you're running in `ts-node-dev`: https://github.com/whitecolor/ts-node-dev/issues/108.

Setting `ts-node.register.instance` was a proposed solution but it seems they just ended up with an environment variable.
Feel free to re-open the discussion over at their repository to make `ts-node` and `ts-node-dev` more uniform. 🙂

jmfury:
> I was confused as well and did some digging. They only recently added the environment variable option as a solution to checking if you're running in ts-node-dev: wclr/ts-node-dev#108.

:wave:  Just wanted to point out an issue I had getting going (with project using `ts-node-dev`).

The errors I got from `fastify-autoload` while using `ts-node-dev@1.0.0-pre.50` sort of made me question the types & my own module resolution. This was not the case.

Turns out the fix was simply to bump the version to `ts-node-dev@1.1.1`  since it was the *environment variable* missing from this package in that prerelease. 

I think the check for the `process.env.TS_NODE_DEV` is all good, just a rough patch with this dependency. 

I don't think I have the context to contribute a docs/README entry but _minimum versions_ of the dev deps might be good for any future Typescript notes. 

Thought I'd share in case it helps anyone else down the road :smile:

mcollina:
thanks for sharing!

PR #107, opened 2020-09-15
Synchronize (.github) configurations by ghconfig

StarpTech:


jsumners (commented):
Is this adding files that are supposed to be provided by `github.com/fastify/.github`?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,38 @@
...
+    paths-ignore:
+    - '**.md'
+    - docs/**
+  pull_request:
+    paths-ignore:
+    - '**.md'
+    - docs/**
+env:
+  A: 40317225-282a-425b-8fe6-1656f182fcc1
+  B: fastify/fastify-autoload
```
mcollina, line 13: what are A and B? can we get more distict names?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,38 @@
...
+  CI: "true"
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        node-version:
+        - 12.x
```
mcollina, line 21: This is not correct, we support Node 10.x as well.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,38 @@
...
+    - uses: actions/checkout@v2
+    - uses: actions/setup-node@v1
+      name: Use Node.js ${{ matrix.node-version }}
+      with:
+        node-version: ${{ matrix.node-version }}
+    - name: install
+      run: |
+        yarn install
+    - name: test
+      run: yarn test
```
mcollina, line 36: This should need npm.

.github/workflows/release.yaml
```diff
@@ -0,0 +1,38 @@
...
+      name: Use Node.js ${{ matrix.node-version }}
+      with:
+        node-version: ${{ matrix.node-version }}
+    - name: install
+      run: |
+        yarn install
+    - name: test
+      run: yarn test
+    needs:
+    - a
```
mcollina, line 38: This seems non-relevant

.github/workflows/cron.yaml
```diff
@@ -0,0 +1,37 @@
...
+      with:
+        node-version: ${{ matrix.node-version }}
+    - name: install
+      run: |
+        yarn install
+    - name: test
+      run: yarn test
+  foo:
+    container:
+      image: nginx
```
mcollina, line 37: This does nto seems relevant

mcollina (changes requested):
1. A lot of those files are not needed.
2. this should remove the old ".dependabot" folder

StarpTech:
This should just test the tool 😄

PR #109, opened 2020-10-01
Add maxDepth on plugin options type interface

aryraditya:
Add missing maxDepth property on Autoload Plugin Options Interface


mcollina (changes requested):
Can you please add a unit test?

mcollina (approved):
lgtm

PR #126, opened 2021-01-19
fix: fixed error when importing using esm syntax

alemagio:
#### Checklist

When importing autoload using `import autoload from 'fastify-autoload'` I get an error which `Error: plugin must be a function or a promise`.
Using commonjs syntax works fine but I believe we want to make it work in both scenarios.

I changed the unit test which was using `require` to include autoload.

PS
I also believe this might be the fix needed [here](https://github.com/fastify/fastify-cli/issues/303) too.

### DIFF
--- a/test/typescript/basic/app.ts
+++ b/test/typescript/basic/app.ts
@@ -1,6 +1,6 @@
 import { FastifyPlugin } from 'fastify'
 import { join } from 'path'
-const fastifyAutoLoad = require('../../../')
+import fastifyAutoLoad from '../../../'
 
 const app: FastifyPlugin = function (fastify, opts, next): void {
   fastify.register(fastifyAutoLoad, {
--- a/fastify-autoload.d.ts
+++ b/fastify-autoload.d.ts
@@ -18,3 +18,4 @@ declare const fastifyAutoload: FastifyPlugin<AutoloadPluginOptions>
 type RewritePrefix = (folderParent: string, folderName: string) => string | boolean
 
 export default fastifyAutoload
+export { fastifyAutoload }
--- a/index.js
+++ b/index.js
@@ -19,7 +19,7 @@ const defaults = {
   dirNameRoutePrefix: true
 }
 
-module.exports = async function fastifyAutoload (fastify, options) {
+const autoload = async function fastifyAutoload (fastify, options) {
   const packageType = await getPackageType(options.dir)
   const opts = { ...defaults, packageType, ...options }
   const plugins = await findPlugins(opts.dir, opts)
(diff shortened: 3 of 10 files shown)


RafaelGSS (commented):
Can you add unit tests for esm?

test/typescript/basic/app.ts
```diff
@@ -1,6 +1,6 @@
 import { FastifyPlugin } from 'fastify'
 import { join } from 'path'
-const fastifyAutoLoad = require('../../../')
+import fastifyAutoLoad from '../../../'
```
kibertoad, line 3: Full set of import tests would be nice, see https://github.com/fastify/fastify-funky/blob/main/index.test-d.ts for an example.

mcollina (approved):
lgtm

alemagio:
@RafaelGSS 
I pushed an update, I think esm cases should be covered by existing tests and the one suggested by @kibertoad 
But if you meant something different just let me know.

kibertoad (approved):

RafaelGSS:
@alemagio Not sure if ESM cases are covered by typescript tests. I was thinking simple tests like it: https://github.com/fastify/fastify/blob/master/test/esm/esm.mjs#L4

alemagio:
> @alemagio Not sure if ESM cases are covered by typescript tests. I was thinking simple tests like it: https://github.com/fastify/fastify/blob/master/test/esm/esm.mjs#L4

Please correct me if I'm wrong but maybe that case is already covered by https://github.com/fastify/fastify-autoload/blob/master/test/module/basic/foo/module.mjs
What do you think @RafaelGSS ?

Anyway I added some more tests using all the possible esm import syntax. (tests are never enough :joy: )

dimitar-cp:
Hi, 
At my company we're considering using fastify and typescript for a project. 
Is this going to be merged any time soon?
Thanks

kibertoad:
@RafaelGSS Were your concerns addressed fully?

kibertoad:
@dimitar-cp This issue is in no way a blocker for adoption of fastify with typescript, more of an annoyance that can be worked around.

RafaelGSS (approved):
LGTM

dimitar-cp:
> @dimitar-cp This issue is in no way a blocker for adoption of fastify with typescript, more of an annoyance that can be worked around.

Thanks, @kibertoad , 
You're right in general of course, but one of the things we discussed when evaluating different frameworks was the availability of generators and this one just didn't work. I am really happy about the quick response though. Gives me leverage to push forward to fastify adoption.
Thanks again 👍🏻

mcollina:
Shipped as v3.4.1

PR #127, opened 2021-01-25
fix: skip override on exports

simoneb:
This fixes an issue introduced in 3.4.1 whereby skip override was only set on the cjs export.

It would make the esm import fail like here: https://github.com/nearform/the-fastify-workshop/pull/12

Note: I had to fix plenty of linting errors because the file I spitted into multiple modules was ignored


index.js
```diff
@@ -211,7 +211,7 @@ function enrichError (err) {
 
 // do not create a new context, do not encapsulate
 // same as fastify-plugin
-module.exports[Symbol.for('skip-override')] = true
+fastifyAutoload[Symbol.for('skip-override')] = true
```
simoneb, line 214: this is the only actual change

Issue #79, opened 2020-06-03
How to apply a lifecycle hook to all plugins in a folder


mcollina:
It would be really handy to be able to set a few lifecycle hooks for each folder. This would be really handy to add authentication for a specific folder folder for example.

iainjreid:
Hi @mcollina, would this be a good first issue? If so, is this the sort of interface you were thinking of?

```javascript
app.register(autoload, {
  dir: path.join(__dirname, 'plugins')
  hooks: {
    onRequest: [auth]
  }
})
```

mcollina:
I was thinking more of a .autoload.js file that is automatically loaded with fastify-plugin for a given folder.

iainjreid:
Something like this?

```javascript
// .autoload.js
module.exports = {
  onRequest: [auth]
}
```

Should developers be able to have a different `.autoload.js` files in each plugin directory where needed (replacing the hooks inherited from the parent directory)?

mcollina:
I would go and make .autoload.js a full plugin. Yes, those should be nested.

iainjreid:
I'm really sorry, but I've looked through the codebase and I'm not sure how to go about this, could you point me in the right direction?

I can see how adding an extra block of code to the `fastifyAutoload` function could load an `.autoload.js` file and use the contents to mutate each plugin found in the child directories upon registration.

mcollina:
Roughly given this tree:

```
/routes/a/
  |- .autoload.js
  |- b.js
```


Note that in order to achieve encapsulation you would need to always create a new plugin in https://github.com/fastify/fastify-autoload/blob/master/index.js#L79, and then call then call https://github.com/fastify/fastify-autoload/blob/dd3f3312bd8fbf01cbf7b4375932a5fbf2aeb55f/index.js#L157 on this plugin. You would need to reset the `prefix` when doing this.

After doing this, you just register `autoload.js` on this intermedia plugin.

kswope:
Just starting looking at a minimalist JS framework for a larger project and fastify looks like its almost a perfect fit but this particular issue is a bit of a show stopper for me and may force me up to less minimal framework.

The autoload plugin is really a killer feature for a minimal framework (Its strange I had to find it in the starter template, its never mentioned in the docs at all!)

But the problem, brought up in this issue, is that there’s no easy way to scope a plugin to a “branch” of the routes.

I came up with a ‘handoff’ strategy but it really seems like the plugin is almost there.

[...]

PR #128, opened 2021-01-31
feat: autoload hooks

polymathca:
Adds automatic insertion of hooks from `.autohook.js` files in route folders, with options to cascade hooks to children and allow subsequent `.autohook.js` files to restart the cascade. See #79 

#### Checklist


index.js
```diff
@@ -16,17 +16,28 @@ const moduleSupport = semver.satisfies(process.version, '>= 14 || >= 12.17.0 < 1
...
   const packageType = await getPackageType(options.dir)
   const opts = { ...defaults, packageType, ...options }
   const plugins = await findPlugins(opts.dir, opts)
   const pluginsMeta = {}
 
-  await Promise.all(plugins.map(({ file, type, prefix }) => {
-    return loadPlugin(file, type, prefix, opts)
+  try {
+    // bring in fp so we can expose the hooks to the parent plugin
+    if (opts.autoHooks) fp = require('fastify-plugin')
```
mcollina, line 33: just require it at the top. Anyway, it's not strictly needed. The only neeeded is https://github.com/fastify/fastify-plugin/blob/52cec31e4c0dc4eb6ecc2b2f0e2f1c21994713d1/plugin.js#L21.

mcollina (commented):
Looking at the code now, it might be simpler to leverage https://github.com/fastify/fastify/blob/master/docs/Encapsulation.md and add an intermediate parent when the hooks are present. This could remove some of those options and have a simpler behavior.

Wdyt?

(This also needs tests).

index.js
polymathca: > 
> 
> Looking at the code now, it might be simpler to leverage https://github.com/fastify/fastify/blob/master/docs/Encapsulation.md and add an intermediate parent when the hooks are present.

I think this is how I handled it. When there are hooks, it will create an intermediate parent (`fastify.register...`) and register the hook(s) and plugin within that scope (`app.register...`). Otherwise it just registers the plugin at the `fastify.register` level. Did I get some syntax wrong?

> This could remove some of those options and have a simpler behavior.
> 
> Wdyt?
> 

I agree, the first draft of the examples in the documentation make it feel confusing. 

My reasoning behind the two additional hook attachment options was that it felt most logical to cascade by default as that's the way folder permissions tend to work in most operating systems (i.e. if you own a parent folder, and you create subfolders, the ownership cascades until you override it). Same with the default ``.htaccess``-style basic auth (i.e. the parent folder and all the children require the same username and password until you override it with a new .htaccess file)

That would be useful for cases like a news site with a paywall that has different membership levels, or a multi-tenant application with differing feature tiers.

E.g.

  ```
  ├── routes
  │   ├── paywall
  │   │   ├── .autohooks.js // check membership
  │   │   ├── routes.js
  │   │   └── archive
  │   │       └── archive-routes.js // membership check inherited from parent
  │   │           └── premium_archive
  │   │               ├── .autohooks.js // premium membership check
  │   │               └── routes.js
  │   └── public_routes.js
  └── app.js
  ```

But I could also see situations where just one or two folders inside a tree could need a set of hooks (e.g. cache headers).

Maybe I could rename the options to make them more logical?

> just require it at the top. Anyway, it's not strictly needed. 

Done. Much cleaner!

> The only neeeded is https://github.com/fastify/fastify-plugin/blob/52cec31e4c0dc4eb6ecc2b2f0e2f1c21994713d1/plugin.js#L21.

The main Fastify `Plugin.md` suggested that using fp was the best practice to future-proof the plugin. I wasn't excited about adding a dependency, but I didn't want to add anything that might break in a core update down the road.

> (This also needs tests).

For sure. I started my first project using Fastify last week (and absolutely love the flexibility and speed) so I'm still getting familiar with the code.

index.js
```diff
@@ -184,7 +235,21 @@ function registerPlugin (fastify, meta, allPlugins, parentPlugins = {}) {
     }
   }
 
-  fastify.register(plugin, options)
+  if (hooks && hooks.length > 0) {
+    // plugin has hooks, which we need to encapsulate at the same layer as the plugin
+    fastify.register(async function (app, opts) {
+      for (const hook of hooks) {
+        if (hook.default) hook.default[Symbol.for('skip-override')] = true
```
mcollina, line 242: You would need to set this to `hook` as well if it's a function.

index.js
```diff
@@ -184,7 +235,21 @@ function registerPlugin (fastify, meta, allPlugins, parentPlugins = {}) {
...
+        app.register(hook, options)
+      };
+
+      app.register(plugin, options)
+    })
```
mcollina, line 247: I think we should invert this relationship as it can be inefficient as it would load them for all files. I recommend we create a new plugin, load all hooks in a folder, and then we call `registerPlugin` for all the routes files.

mcollina (changes requested):
Good work! You'd need to add tests for commonjs as well

index.js
polymathca: Sounds good, I moved that logic to a wrapper function and now pass in a completed plugin object for the `registerPlugin` call. Does that make more sense to you?

mcollina (changes requested):
After reading up in #79 and the discussion about hidden files, we might consider using a different way to differentiate the autohooks instead of `.autohooks.mjs`. I think we might just go with `autohooks.js` or even `hooks.js`.

index.js
mcollina: That's not what I'm looking for close.

In https://github.com/fastify/fastify-autoload/blob/fa95893d3d5fc954bece8ffd8d3cd974d7154d78/index.js#L43, we area loading all plugins in a loop. We are flattening a tree into a list. This will cause us executing the same hooks for each file in the in the given sub-tree.

The most efficient way of loading this would be to retain the information and just load them once (in pseudocode):

```js
    // plugin has hooks, which we need to encapsulate at the same layer as the plugin
    fastify.register(async function (app, opts) {
      app.register(hooks)
      for (const plugin of pluginsInThisFolder) {
        app.register(plugin, options)
      }
    })
```

polymathca:
> 
> 
> After reading up in #79 and the discussion about hidden files, we might consider using a different way to differentiate the autohooks instead of `.autohooks.mjs`. I think we might just go with `autohooks.js` or even `hooks.js`.

Agreed! My vote would be default to `autohooks.js` so it still gets close to the top of the folder listing in an IDE, and still signals that there's some magic happening. I recommend that we keep the regex for that exposed as an option (I added a line to the documentation) so people can adapt it to their preferences.

index.js
polymathca: That makes a lot of sense, I will take a look tonight!

mcollina:
I agree.

polymathca:
I think there's room for a bit of refactoring here, but I wanted to see if it falls in line with what you were thinking.

The flat plugins array in the `findPlugins` function is behaving as before, but I've added a second where both the hooks and plugins are inserted into an object using their prefix (or /) as keys.

Then I loop through the prefixed object, load the hooks, create a plugin function, register the plugins at the `app` level within the new plugin to keep them encapsulated, and the main plugins at the `fastify` level.

Do you think this prefixed object strategy is adding too much additional complexity? I don't love having two objects coming out of the `findPlugins` function with such similar data, but that is the most logical place to build the map of route prefixes and determine where hooks need to be attached.

Let me know what you think!

README.md
```diff
@@ -50,6 +50,13 @@ Folder structure:
 
 ```
 ├── plugins
+│   ├── hooked-plugin
+│   │   ├── .autohooks.mjs
```
mcollina, line 54:
```suggestion
│   │   ├── autohooks.mjs
```

index.js
```diff
@@ -37,9 +38,25 @@ const fastifyAutoload = async function autoload (fastify, options) {
       })
   }))
 
-  for (const name in pluginsMeta) {
-    const plugin = pluginsMeta[name]
-    registerPlugin(fastify, plugin, pluginsMeta)
+  if (!options.autoHooks) { // default behaviour: apply plugins at root level of fastify app, no encapsulation required
```
mcollina, line 41: I think we might remove this branch altogether.

mcollina (changes requested):
Could you please add support for:

```
│   ├── hooked-plugin
│   │   ├── autohooks.mjs
│   │   ├── routes.js
│   │   └── children
│   │       ├── autohooks.mjs
│   │       ├── commonjs.cjs
│   │       ├── module.mjs
│   │       └── typescript.ts
```

With the behavior of both hooks file should be reentrantly loaded

I don't think that's supported right now.

index.js
polymathca: I agree, that will make things a lot easier to follow.

I will refactor this section and add a few benchmarks to make sure this isn't adding any significant overhead.

polymathca:
> 
> 
> Could you please add support for:
> 
> ```
> │   ├── hooked-plugin
> │   │   ├── autohooks.mjs
> │   │   ├── routes.js
> │   │   └── children
> │   │       ├── autohooks.mjs
> │   │       ├── commonjs.cjs
> │   │       ├── module.mjs
> │   │       └── typescript.ts
> ```
> 
> With the behavior of both hooks file should be reentrantly loaded
> 
> I don't think that's supported right now.

I believe this revision covers that now, as well, but I'm open to suggestions!

mcollina:
This seems good to go! Is there anything missing from your point of view?

README.md
```diff
@@ -161,6 +168,52 @@ Autoload can be customised using the following options:
   // routes can now be added to /defaultPrefix/something
   ```
 
+- `autoHooks` (optional) - Apply hooks from `autohooks.js` file(s) to plugins found in folder
```
jsumners, line 171: The above tree shows `autohooks.mjs` but this documentation does not mention ESM support.

README.md
```diff
@@ -161,6 +168,52 @@ Autoload can be customised using the following options:
...
+  fastify.register(autoLoad, {
+    dir: path.join(__dirname, 'plugins'),
+    autoHooks: true,
+    autoHooksPattern: /^[_.]?auto_?hooks(\.ts|\.js|\.cjs|\.mjs)$/i
+  })
+  ```
+
+- `cascadeHooks` (optional) - If using `autoHooks`, cascade hooks to all children. Ignored if `autoHooks` is `false`.
+
+  Default behaviour of `autoHooks` is to apply hooks only to the level on which the `autohooks.js` file is found. Setting `cascadeHooks: true` will continue applying the hooks to any children.
```
jsumners, line 194: Is this wrapping the hook in `fastify-plugin`?

README.md
polymathca: Good catch, I revised the wording and example in the autoHooks section to reflect both.

README.md
polymathca: No, it's adding `skip-override: true` to the hook plugin(s) and encapsulating them with the routes, which are added to the `fastify` / `app` instance where the `fastify-autoload` plugin is registered.

polymathca:
> 
> 
> This seems good to go! Is there anything missing from your point of view?

I think it's looking pretty good, I've been using it on a project for the past week and haven't encountered any surprises!

Is the test coverage adequate?

README.md
jsumners: A note about this being implemented via Fastify's encapsulation feature and a link to that documentation would be good.

polymathca: Sounds good, which is the preferred link to the documentation?

https://www.fastify.io/docs/latest/Encapsulation/

or 

https://github.com/fastify/fastify/blob/master/docs/Encapsulation.md

jsumners: I'd go with the first one.

polymathca: I changed the wording to reflect that it's encapsulating files and added the link, does that work a bit better?

I also added a `routes.js` to the example tree in the autoHooks section, so it was more obvious that it was a plugin being loaded by `fastify-autoload` but unaffected by the hooks.

climba03003:
It is better to add unit test for the line of code you add.
Missing line `54,135,138,290`.

If you think this PR is ready. You should change it as ready for review.

jsumners (approved):
LGTM

mcollina (approved):
lgtm

Issue #133, opened 2021-02-22
Support route parameters in folder names 


mcollina:
## 🚀 Feature Proposal

We should support automatic routes for route params, as example `/user/:userId/cart`

```
├── routes
│   ├── user
│   │   ├── _userId
│   │       └── cart.js
├── package.json
└── app.js
```

I would recommend we convert `_` to `:` on loading.

zekth:
I like this implementation and a good feature would be the generation of the scaffolding using `fastify-cli`.

jsumners:
With as many customizations this module is getting for hooking into the algorithm, it might just be better to require an `autoload.{json,yaml}` file that has mappings and other configurations.

mcollina:
I tend to agree?

polymathca:
> 
> 
> With as many customizations this module is getting for hooking into the algorithm, it might just be better to require an `autoload.{json,yaml}` file that has mappings and other configurations.

Were you thinking it would live in the root of the folder from which plugins are being loaded, or in a separate config folder elsewhere in the project?

jsumners:
> > With as many customizations this module is getting for hooking into the algorithm, it might just be better to require an `autoload.{json,yaml}` file that has mappings and other configurations.
> 
> Were you thinking it would live in the root of the folder from which plugins are being loaded, or in a separate config folder elsewhere in the project?

I imagine it would be a file within each directory that is loaded:

```
/project
  \__ /plugins
          \__ /plugin1
                 \__ autoload.yaml
```

🤷‍♂️  I'm just making an observation. It could be an unworkable idea or not worth the effort. I really do not know.

delvedor:
I like using the folder names for generating the path, it's a well-established pattern in the node community.
Something like:
```

├── routes
│   ├── user
│   │   ├── [userId]
│   │       └── cart.js
├── package.json
└── app.js
```

Configuration files could work as well, but yaml are very easy to misconfigure. Furthermore, at that point, you could just use the `autoPrefix` export 🤔 .

fox1t:
I like the idea of using the filesystem since it is a pattern popularized by Next.js. I don't like having a separate folder for routes: usually, I divide the code by "services ."I think that using a `yml` provides more flexibility, though. 

We might support both, starting from the filesystem thing?

mcollina:
FYI, I think the yml/json file is orthogonal to this issue, maybe we can split it?

polymathca:
> 
> 
[...]

PR #134, opened 2021-02-23
feat: route parameters

polymathca:
Adds route parameters based on underscore-prefixed folders. See #133 

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -222,6 +222,31 @@ Autoload can be customised using the following options:
   })
   ```
 
+- `routeParams` (optional) - Folders prefixed with `_` will be turned into route parameters.
+
+  ```js
+  /*
+  ├── routes
+  │   └── users
+  │       ├── _id
+  │       │   └── actions.js
+  │       └── index.js
+  └── app.js
+  */
+
+  fastify.register(autoLoad, {
+    dir: path.join(__dirname, 'routes'),
+    routeParams: true // routes/users/_id/actions.js will be loaded with prefix /users/:id
+  })  
+
+  // curl http://localhost:3000/users/index
+  // { userIndex: [ { id: 7, username: 'example' } ] }
+
+  // curl http://localhost:3000/users/7/details
+  // { user: { id: 7, username: 'example' } }
+
+  ```
+
 ## Plugin Configuration
 
 Each plugin can be individually configured using the following module properties:
--- a/index.js
+++ b/index.js
@@ -12,12 +12,12 @@ const hasTsJest = 'npm_package_devDependencies_ts_jest' in process.env
 const typescriptSupport = isTsNode || (isJestEnviroment && hasTsJest)
 
 const moduleSupport = semver.satisfies(process.version, '>= 14 || >= 12.17.0 < 13.0.0')
+const routeParamPattern = /\/_/i
 
 const defaults = {
   scriptPattern: /((^.?|\.[^d]|[^.]d|[^.][^d])\.ts|\.js|\.cjs|\.mjs)$/i,
   indexPattern: /^index(\.ts|\.js|\.cjs|\.mjs)$/i,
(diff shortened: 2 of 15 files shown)


polymathca:
I just noticed some strange behaviour in the output of `fastify.printRoutes()` if a file inside the parameterized folder (e.g. `_id`) doesn't have a root path defined (`/`), but does have a named path (`/test`) defined.

Given

```js

# /routes/_id/index.js

export default async function (app, opts) {
  app.get('/test', async function (req, reply) {
  })
}

```

Instead of showing:

```
├── :id/test (GET)
└── / (GET)
```

it displays:

```
├── :/testid (GET)
└── / (GET)
```


The application itself still functions as expected, e.g. `/routes/some_id/test` will respond correctly. Adding;

```js
app.get('/', async function (req, reply) {
})
```

Results in:

```
├── :id (GET)
└── / (GET)
    └── test (GET)
        └── / (GET)
```

Not sure what to make of it!

climba03003:
My advice would be make use of `fastify.log.debug` or `fastify.log.trace` when you debug.
Then, just left the the `log` there as they are useful when you debugging. And this info should be useful when the others want to debug.

polymathca:
> 
> 
> My advice would be make use of `fastify.log.debug` or `fastify.log.trace` when you debug.
> Then, just left the the `log` there as they are useful when you debugging. And this info should be useful when the others want to debug.

Thanks, I'll keep that in mind! I wasn't sure where to start logging, as the route is working properly, just not displaying correctly in the `printRoutes()` output.

polymathca:
I added a new test to track this down. The route tree, at least on my system, looks like this:

```
└── / (GET)
    ├── pages (GET)
    │   └── / (GET)
    │       ├── archived (GET)
    │       └── :id (GET)
    │           └── / (GET)
    │               └── edit (GET)
    └── users/:/detailsid (GET)   <- this is scrambled, should be 'users/:id/details (GET)'
```

But all of the tests pass as expected, so it appears to just be an issue in how the tree is being displayed.

mcollina:
The code looks good to me.
It would be awesome if you could identify why find-my-way printRoutes scrambles it (and maybe send a PR).

I would start trying to assemble the routes manually and see if you can reproduce the rendering problem.

polymathca:
> 
> 
> The code looks good to me.
> It would be awesome if you could identify why find-my-way printRoutes scrambles it (and maybe send a PR).
> 
> I would start trying to assemble the routes manually and see if you can reproduce the rendering problem.

I believe I finally found the culprit! I created a PR here:

https://github.com/delvedor/find-my-way/pull/182

(I'm not sure if github allows referencing between projects or not)

mcollina:
The code looks ok! Could you add docs about it in the README.md?

polymathca:
I updated the documentation and added CJS tests as well, should be good to go!

README.md
```diff
@@ -222,6 +222,31 @@ Autoload can be customised using the following options:
   })
   ```
 
+- `routeParams` (optional) - Folders prefixed with `_` will be turned into route parameters.
```
mcollina, line 225: There is also `routeParamPattern` to document

mcollina (changes requested):
lgtm

README.md
polymathca: Is there an example of a __warning: this could break things__ note elsewhere in the Fastify documentation?

If we expose this as a configuration option, I think it's important to point out that if this is set improperly it will cause all sorts of headaches (e.g. forgetting `^` and suddenly all of your routes are parametric)

mcollina: Stating it like you said it's ok.

I'm also ok in not making it configurable - however it's configurable in the code.

polymathca: Good call, I think it's better to move it out of the configuration scope for simplicity's sake. Prefixing with `_` is a good convention, and I think anything more complex is probably better served in application code instead of the filesystem.

mcollina (approved):
lgtm

PR #144, opened 2021-03-09
feat: Added forceESM option.

ShogunPanda:
This PR adds the `forceESM` option, which always uses `await import` to load modules despite of the detected package type.

I run into this when running code via `ts-node` in a folder where `package.json` had `type='module'`.

I also tried to add a test for this, but I was not able to reproduce the problem in this repo, since `type!='module'`.

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,6 +139,15 @@ Autoload can be customised using the following options:
   })
   ```
 
+- `forceESM` (optional) - If set to 'true' it always use `await import` to load plugins or hooks.
+
+  ```js
+  fastify.register(autoLoad, {
+    dir: path.join(__dirname, 'plugins'),
+    forceESM: true
+  })
+  ```
+
 - `options` (optional) - Global options object used for all registered plugins
 
   Any option specified here will override `plugin.autoConfig` options specified in the plugin itself.
--- a/fastify-autoload.d.ts
+++ b/fastify-autoload.d.ts
@@ -11,6 +11,7 @@ export interface AutoloadPluginOptions {
   indexPattern?: RegExp
   options?: FastifyPluginOptions
   maxDepth?: number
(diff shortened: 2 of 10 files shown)


mcollina (changes requested):
Could you add a unit test for this new functionality? It's not just a definition change.

ShogunPanda:
I already tried above but I was not able to reproduce the problem within the repo.

I've tried again with several strategies. Unfortunately I haven't succeeded.
What shall we do?

ShogunPanda:
Done!

climba03003:
Can you fix the failing CI?

ShogunPanda:
Sure, it's fixed now!

RafaelGSS:
What problem this solves? Reading the description seems to me that `fastify-autoload` doesn't support esm yet.

> I run into this when running code via ts-node in a folder where package.json had type='module'.

 That's it?

ShogunPanda:
No, it's actually a (not sure how much) edge case with TypeScript.

If you use `ts-node` in a folder whose applicable `package.json` has `type=module`, fastify-autoload will fail.

This is because of the following stesp: 

1. `https://github.com/fastify/fastify-autoload/blob/master/index.js#L98` return `typescript`.
2. `https://github.com/fastify/fastify-autoload/blob/master/index.js#L216` will therefore attempt to load using `require`.
3. Since `type=module`, `.js` (or `.ts`) files are interpreted by default from Node as ESM and thus `require` shall not be used and will fail.

My PR fixes step 2 by adding a `forceESM` that bypasses detection in step 1.

RafaelGSS:
Ok, thanks!

RafaelGSS (approved):
LGTM.

mcollina (approved):
lgtm

PR #216, opened 2022-01-08
chore: testing new reusable workflow

Eomm:
Testing the new reusable workflow

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,4 +1,4 @@
-name: CI
+name: ci
 
 on:
   push:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,6 +12,4 @@ on:
 
 jobs:
   test:
-    uses: fastify/workflows/.github/workflows/plugins-ci.yml@v1
-    with:
-      generate-coverage: true
+    uses: fastify/workflows/.github/workflows/plugins-ci.yml@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,7 +11,7 @@ on:
       - '*.md'
 
 jobs:
-  call-reuseable-workflow:
+  test:
     uses: fastify/workflows/.github/workflows/plugins-ci.yml@v1
     with:
       generate-coverage: true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,45 +1,17 @@
-name: ci
-'on':
+name: CI
+
+on:
   push:
     paths-ignore:
-      - docs/**
+      - 'docs/**'
       - '*.md'
   pull_request:
     paths-ignore:
-      - docs/**
+      - 'docs/**'
       - '*.md'
+
 jobs:
-  build:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -1,45 +1,17 @@
...
-    needs: build
-    runs-on: ubuntu-latest
-    permissions:
-      pull-requests: write
-      contents: write
-    steps:
-      - uses: fastify/github-action-merge-dependabot@v3
-        with:
-          github-token: ${{ secrets.GITHUB_TOKEN }}
+  call-reuseable-workflow:
```
mcollina, line 14:
```suggestion
  test:
```

mcollina (commented):
Seem to work great! Just a note!

Eomm:
There was an error on the coverage step that requires the `--cov --coverage-report=html` args.

```
Run coverallsapp/github-action@master
  with:
    github-token: ***
    parallel: true
    flag-name: run-12-ubuntu-latest
    path-to-lcov: ./coverage/lcov.info
    coveralls-endpoint: https://coveralls.io
Using lcov file: ./coverage/lcov.info
Error: Lcov file not found.
```

Now the error is the following:

```
Run coverallsapp/github-action@master
Using lcov file: ./coverage/lcov.info
Error: Bad response: 422 {"message":"Couldn't find a repository matching this job.","error":true}
```

Not sure the config required to get the coverrals working.

mcollina:
I would just remove coveralls from  everywhere.

Fdawgs:
> 
> Now the error is the following:
> 
> 
> 
> ```
> 
> Run coverallsapp/github-action@master
> 
> Using lcov file: ./coverage/lcov.info
> 
> Error: Bad response: 422 {"message":"Couldn't find a repository matching this job.","error":true}
> 
> ```
> 
> 
> 
> Not sure the config required to get the coverrals working.
> 
> 

I think I talked about this with @mcollina early last year back when I was trying to standardise the plug-in repos. Some of the repos in the Fastify org do not send coverage to Coveralls or fail to do so if they try (like this one). Fastify-cors is an example of one that does manage to send coverage.

mcollina:
I would just disable it everywhere, it's not worth it (effort vs the information it provides).

zekth:
+1 on disabling coveralls if it's a blocker for this feature

Fdawgs:
@Eomm, v2 of workflows has been released, which has removed all coveralls integration.

.github/workflows/ci.yml
```diff
@@ -1,45 +1,15 @@
-name: ci
-'on':
+name: CI
```
Fdawgs, line 1: The CI badge in the readme may need to be updated if changing from lower to uppercase, but i prefer the uppercase to be honest!

Fdawgs (approved):

.github/workflows/ci.yml
Eomm: Good spot: for consistency in the action menu, I would keep the old `ci` name.

I must say that I like the upper case too 😖

Fdawgs:
LGTM 👍

PR #222, opened 2022-02-08
feat: adding `encapsulate` option with default `true`, non-breaking change

wilbert-abreu:
Adding the option below:

`encapsulate` (optional) - Defaults to `true`, if set to `false` each plugin loaded is wrapped with [fastify-plugin](https://github.com/fastify/fastify-plugin). This allows you to share contexts between plugins and the parent context if needed. For example, if you need to share decorators. Read [this](https://github.com/fastify/fastify/blob/main/docs/Reference/Encapsulation.md#sharing-between-contexts) for more details.

  ```js
  fastify.register(autoLoad, {
    dir: path.join(__dirname, 'plugins'),
    encapsulate: false
  })
  ```

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -146,6 +146,14 @@ Autoload can be customised using the following options:
     forceESM: true
   })
   ```
+- `encapsulate` (optional) - Defaults to 'true', if set to 'false' each plugin loaded is wrapped with [fastify-plugin](https://github.com/fastify/fastify-plugin). This allows you to share contexts between plugins and the parent context if needed. For example, if you need to share decorators. Read [this](https://github.com/fastify/fastify/blob/main/docs/Reference/Encapsulation.md#sharing-between-contexts) for more details.
+
+  ```js
+  fastify.register(autoLoad, {
+    dir: path.join(__dirname, 'plugins'),
+    encapsulate: false
+  })
+  ```
 
 - `options` (optional) - Global options object used for all registered plugins
 
@@ -192,7 +200,7 @@ Autoload can be customised using the following options:
   })
   ```
 
-  If `autoHooks` is set, all plugins in the folder will be [encapsulated](https://github.com/fastify/fastify/blob/master/docs/Encapsulation.md)
+  If `autoHooks` is set, all plugins in the folder will be [encapsulated](https://github.com/fastify/fastify/blob/main/docs/Reference/Encapsulation.md)
   and decorated values _will not be exported_ outside the folder.
 
 - `autoHooksPattern` (optional) - Regex to override the `autohooks` naming convention
--- a/index.js
+++ b/index.js
@@ -19,7 +19,8 @@ const defaults = {
   scriptPattern: /((^.?|\.[^d]|[^.]d|[^.][^d])\.ts|\.js|\.cjs|\.mjs)$/i,
   indexPattern: /^index(\.ts|\.js|\.cjs|\.mjs)$/i,
   autoHooksPattern: /^[_.]?auto_?hooks(\.ts|\.js|\.cjs|\.mjs)$/i,
-  dirNameRoutePrefix: true
+  dirNameRoutePrefix: true,
+  encapsulate: true
 }
 
 const fastifyAutoload = async function autoload (fastify, options) {
@@ -217,7 +218,7 @@ async function findPlugins (dir, options, hookedAccumulator = {}, prefix, depth 
 }
 
 async function loadPlugin (file, type, directoryPrefix, options) {
-  const { options: overrideConfig, forceESM } = options
+  const { options: overrideConfig, forceESM, encapsulate } = options
   let content
   if (forceESM || type === 'module') {
     content = await import(url.pathToFileURL(file).href)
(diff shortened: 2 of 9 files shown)


wilbert-abreu:
Hey @zekth @darkgl0w @Eomm, when you get a moment, would love some feedback here. Thanks!

README.md
```diff
@@ -146,6 +146,14 @@ Autoload can be customised using the following options:
     forceESM: true
   })
   ```
+- `shareContext` (optional) - Defaults to 'false', if set to 'true' each plugin loaded is wrapped with [fastify-plugin](https://github.com/fastify/fastify-plugin). This allows you to share contexts between plugins and the parent context if needed. For example, if you need to share decorators. Read [this](https://github.com/fastify/fastify/blob/main/docs/Reference/Encapsulation.md#sharing-between-contexts) for more details.
```
mcollina, line 149: I prefer an `encapsulate` flag that defaults to `true`.

mcollina (commented):
Good work!

README.md
wilbert-abreu: that's funny, I did exactly that and called this `encapsulated` at first then changed it since I thought `shareContext` would be more explicit user friendly. If you feel strongly about it, happy to change.

zekth: In the fastify doc we speak about `context` in the encapsulation part with the doc you linked: https://github.com/fastify/fastify/blob/main/docs/Reference/Encapsulation.md#sharing-between-contexts
Mainly the feature is called encapsulation. I join @mcollina on this point, we should uniformize the wording.

package.json
```diff
@@ -6,6 +6,7 @@
   "types": "fastify-autoload.d.ts",
   "scripts": {
     "lint": "standard | snazzy",
+    "lint:fix": "standard --fix | snazzy",
```
zekth, line 9: Noice 👍

zekth (approved):
Except the wording remark good for me.

Thanks!

README.md
wilbert-abreu: Fair, I get that. Will do

wilbert-abreu: @mcollina @zekth done!

package.json
wilbert-abreu: thanks!!

darkgl0w (changes requested):
We need to move `fastify-plugin` to the from the `devDependencies` to the `dependencies`

Otherwise all seems good to me.

wilbert-abreu:
> We need to move `fastify-plugin` from the `devDependencies` to the `dependencies`.
> 
> Otherwise all seems good to me.

oh wow, great catch! Will do

darkgl0w (approved):
lgtm

zekth (approved):

wilbert-abreu:
Thanks all, appreciate all the feedback. Can someone merge and publish when they can? Thanks again!

wilbert-abreu:
Thanks @zekth @darkgl0w @mcollina, appreciate all the feedback. Can someone merge and publish when they can? Thanks again!

README.md
```diff
@@ -146,6 +146,14 @@ Autoload can be customised using the following options:
     forceESM: true
   })
   ```
+- `encapsulate` (optional) - Defaults to 'true', if set to 'false' each plugin loaded is wrapped with [fastify-plugin](https://github.com/fastify/fastify-plugin). This allows you to share contexts between plugins and the parent context if needed. For example, if you need to share decorators. Read [this](https://github.com/fastify/fastify/blob/main/docs/Reference/Encapsulation.md#sharing-between-contexts) for more details.
```
Eomm, line 149: Would this option be like all the plugins are in the same context?
For example this directory tree:

```
/foo
 - foo.js
/bar
 - bar.js
 - baz.js
```

All the *.js files are loaded in the same context?

Would not be more useful to create 2 contexts in this case: one for the `foo/` dir and the `bar/` one?

index.js
```diff
@@ -29,6 +31,9 @@ const fastifyAutoload = async function autoload (fastify, options) {
   const pluginsMeta = {}
   const hooksMeta = {}
 
+  const fastifyRegister = fastify.register
+  fastify.register = (plugin, options) => opts.encapsulate ? fastifyRegister(plugin, options) : fastifyRegister(fp(plugin), options)
```
Eomm, line 35: What is the `plugin` function is already a plugin?
We should check the `fn[Symbol.for('skip-override')]` value

mcollina (changes requested):
On a second look, I do not think the implementation is correct.

Instead of overriding `register`, you should be setting `plugin[Symbol.for('skip-override')] = true` at https://github.com/fastify/fastify-autoload/blob/ab5f5f2a344dd995eb18e29b7ccbb7c35d441145/index.js#L231

wilbert-abreu:
> On a second look, I do not think the implementation is correct.
> 
> Instead of overriding `register`, you should be setting `plugin[Symbol.for('skip-override')] = true` at
> 
> https://github.com/fastify/fastify-autoload/blob/ab5f5f2a344dd995eb18e29b7ccbb7c35d441145/index.js#L231

TIL thanks for that, way better approach. Done and set fastify-plugin back to a devDependencies.

index.js
wilbert-abreu: thanks @Eomm, addressed as part of @mcollina comment

Eomm (approved):

wilbert-abreu:
Awesome, thanks again for the quick support on this @Eomm @mcollina @zekth @darkgl0w. If I could leave a yelp review, I would say 5/5 stars, great experience! 😁

PR #259, opened 2022-08-06
Add support for @swc-node/register

jclem:
This adds support for [@swc-node/register](https://github.com/swc-project/swc-node), a ts-node alternative from the SWC project.

#### Checklist

  *I ran the tests, but there is no benchmark script.*
- [x] documentation is changed or added *n/a*

Note that although I added a TypeScript unit test file for this change, I could not get the test to *fail*, even before adding my changes to `typescriptSupport`.

An additional question: Would you be open to adding a `FASTIFY_AUTOLOAD_TS` environment variable that can be used to force TypeScript support? Otherwise, I feel people are going to hack things like setting `JEST_WORKER_ID` when they have TypeScript support, but this package isn't aware of it.


mcollina (approved):
lgtm

Issue #271, opened 2022-09-25
babel-node + typescript. Error only with autoloads :  throw new Error(@fastify/autoload cannot import plugin at '${file}'. To fix this error compile TypeScript to JavaScript or use 'ts-node' to run your app.)


Tumulte:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.6.0

### Plugin version

5.4.0

### Node.js version

16.14.2

### Operating system

macOS

### Operating system version (i.e. 20.04, 11.3, 10)

12.6

### Description

I have a rather barebone `server.ts` (below) and I try to run it with typescript, ESM and hot reload.

I've set up a nodemon + babel command line : `  nodemon --exec 'babel-node -x .ts' app.ts` and everything works fine... except @fastify/autoload.

If I rename the autoload files as `.js` it works

I got this error : ` throw new Error(@fastify/autoload cannot import plugin at '${file}'. To fix this error compile TypeScript to JavaScript or use 'ts-node' to run your app.)`

I know there are several other issues with this error, but none have provided a solution. Thanks for your kind help, I can take alternative solutions for ESM+TS+hot reload if you know more compatible wayt

server.ts :
```javascript
import path from 'path'
import AutoLoad from '@fastify/autoload'
import {fastify} from "fastify";

let currentApp: string
const server = fastify()
server.register(require("@fastify/view"), {
    engine: {
        nunjucks: require("nunjucks"),
    },
    root: utils.configPath
});
server.addHook("preHandler", function (request, reply, done) {
    // @ts-ignore
    reply.locals = {
        picaro: path.resolve(__dirname, "index.njk")
    };
    done();
});

// This loads all plugins defined in plugins
// those should be support plugins that are reused
// through your application
server.register(AutoLoad, {
    dir: path.join(__dirname, 'plugins'),
})

// This loads all plugins defined in routes
// define your routes in one of these
server.register(AutoLoad, {
    dir: path.join(__dirname, 'routes'),
})

server.listen({port: 3000})
console.info('Loaded on port 3000')

```

### Steps to Reproduce

set up a basic fastify as typescript with babel-node (config below), launch the command `nodemon --exec 'babel-node -x .ts' app.ts`

```json
{
  "presets": [
    "@babel/preset-env",
    "@babel/preset-typescript"
  ],
  "plugins": [
    "@babel/plugin-proposal-class-properties"
  ]
}
```

[...]

PR #264, opened 2022-08-31
Add support for .ts files when app is not run with ts-node.

kylerush:
This branch adds support for `.ts` files when the application is not run with `ts-node.` The specific use case for this is mentioned in #230. Vitest does not run with `ts-node` (it uses `es-build`) to compile TypeScript. This creates a situation where some`.ts` files that AutoLoad tries to import fails because there is no Typescript transpiler.

This pull request resolves #230.

### To make the tests fail
Replace lines 32-42 in `test/commonjs/error.js` with `app3.register(require('./index-error/app'))`. This will cause the plugin to load `ts-node` from the Node modules and not throw an error to the user that they need to install `ts-node`.

Replace `index.js` with the current version of the file in the `master` branch. This will cause the test files in `test/commonjs/ts-node.js` to error because it cannot load `.ts` files without either being run with ts-node or the ts-node `register` that is added in this branch.

## Notes
I did not modify any documentation as the error messages in `index.js` will still make sense. Happy to revise the documentation if requested.

Per request, I added `vite` and `vitest` as devDependencies and added some vitest tests so that we don't lose vitest support. 

Also happy to make any changes needed to this. Thanks for reviewing!

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin]

PR #272, opened 2022-10-07
271: trying to use babel-node when using typescript files

lasmil:
- when dealing with typescript files, we need to make sure the ts-node is being used to autoload components
- closes: https://github.com/fastify/fastify-autoload/issues/271

### DIFF
--- a/index.js
+++ b/index.js
@@ -162,13 +162,11 @@ async function findPlugins (dir, options, hookedAccumulator = {}, prefix, depth 
   if (indexDirent) {
     const file = path.join(dir, indexDirent.name)
     const type = getScriptType(file, options.packageType)
-    if (type === 'typescript') {
-      if (!typescriptSupport) {
-        try {
-          require('ts-node').register()
-        } catch (e) {
-          throw new Error(`@fastify/autoload cannot import hooks plugin at '${file}'. To fix this error compile TypeScript to JavaScript or use 'ts-node' to run your app.`)
-        }
+    if (type === 'typescript' && !typescriptSupport) {
+      try {
+        require('ts-node').register()
+      } catch (e) {
+        throw new Error(`@fastify/autoload cannot import hooks plugin at '${file}'. To fix this error compile TypeScript to JavaScript or use 'ts-node' to run your app.`)
       }
     }
 
@@ -221,13 +219,11 @@ async function findPlugins (dir, options, hookedAccumulator = {}, prefix, depth 
 
     if (dirent.isFile() && scriptPattern.test(dirent.name)) {
       const type = getScriptType(file, options.packageType)
-      if (type === 'typescript') {
-        if (!typescriptSupport) {
-          try {
-            require('ts-node').register()
-          } catch (e) {
-            throw new Error(`@fastify/autoload cannot import plugin at '${file}'. To fix this error compile TypeScript to JavaScript or use 'ts-node' to run your app.`)
-          }
+      if (type === 'typescript' && !typescriptSupport) {
+        try {
+          require('ts-node').register()
+        } catch (e) {
+          throw new Error(`@fastify/autoload cannot import plugin at '${file}'. To fix this error compile TypeScript to JavaScript or use 'ts-node' to run your app.`)
         }
       }
 
--- a/index.js
+++ b/index.js
@@ -6,7 +6,8 @@ const { readdir } = require('fs').promises
 const pkgUp = require('pkg-up')
 
 const isTsNode = (Symbol.for('ts-node.register.instance') in process) || !!process.env.TS_NODE_DEV
-const isBabelNode = process?.execArgv.some((arg) => arg.indexOf('babel-node') >= 0)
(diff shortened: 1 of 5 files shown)


climba03003 (changes requested):
I don't see any different from https://github.com/fastify/fastify-autoload/pull/264

index.js
```diff
@@ -162,8 +162,14 @@ async function findPlugins (dir, options, hookedAccumulator = {}, prefix, depth
   if (indexDirent) {
     const file = path.join(dir, indexDirent.name)
     const type = getScriptType(file, options.packageType)
-    if (type === 'typescript' && !typescriptSupport) {
-      throw new Error(`@fastify/autoload cannot import hooks plugin at '${file}'. To fix this error compile TypeScript to JavaScript or use 'ts-node' to run your app.`)
+    if (type === 'typescript') {
+      if (!typescriptSupport) {
```
simoneb, line 166: nesting two ifs this way is the same as what it was before, why not just keep it as it was?

lasmil:
> 
There is a difference between the 2 PRs. In this PR I am using:
const type = getScriptType(file, options.packageType)
the type to know if it's 'typescript' or not.

In the other (reverted) PR we are trying to require ts-node even when the type is not 'typescript'.

simoneb:
approach looks reasonable, let's add tests to confirm it works!

climba03003:
> There is a difference between the 2 PRs. In this PR I am using: `const type = getScriptType(file, options.packageType) ` the type to know if it's 'typescript' or not.
> 
> In the other (reverted) PR we are trying to require ts-node even when the type is not 'typescript'.

Still, it didn't convince me why we need to explicitly register `ts-node` for the user.
In the issue, I don't see `ts-node` as devDependency. How can explicitly register `ts-node` can solve the problem of someone that do not install it?

The solution should be detect if the current script run under `babel-node` and allow it as `typescript` runtime.

lasmil:
> 

When I tried with the mentioned project I did indeed needed to install ts-node as a dev dependency. Let me try your solution.

lasmil:
> > 
> 
> When I tried with the mentioned project I did indeed needed to install ts-node as a dev dependency. Let me try your solution.



> > There is a difference between the 2 PRs. In this PR I am using: `const type = getScriptType(file, options.packageType) ` the type to know if it's 'typescript' or not.
> > In the other (reverted) PR we are trying to require ts-node even when the type is not 'typescript'.
> 
> Still, it didn't convince me why we need to explicitly register `ts-node` for the user. In the issue, I don't see `ts-node` as devDependency. How can explicitly register `ts-node` can solve the problem of someone that do not install it?
> 
> The solution should be detect if the current script run under `babel-node` and allow it as `typescript` runtime.

I have pushed a commit to this PR. I tried to find a better way to detect if the script is running under babel-node but unfortunately I was unable to find it.

test/typescript-jest/babel-node/index.test.ts
```diff
@@ -0,0 +1,31 @@
...
+    app.ready(done)
+  })
+
+  it("tests the root route", async function () {
+    const response = await app.inject({
+      method: 'GET',
+      url: '/'
+    })
+    expect(response.statusCode).toEqual(200)
+    expect(JSON.parse(response.payload)).toEqual({ hello: 'world' })
```
simoneb, line 21:
```suggestion
    expect(response.json()).toEqual({ hello: 'world' })
```

test/typescript-jest/babel-node/index.test.ts
```diff
@@ -0,0 +1,31 @@
...
+
+describe("load typescript using babel-node", () => {
+  const app = fastify()
+
+  beforeAll( done => {
+    app.register(AutoLoad, {
+      dir: join(__dirname, '../../commonjs/babel-node/routes')
+    })
+    app.ready(done)
+  })
```
simoneb, line 13: nit: let's use newer async/await syntax

test/commonjs/babel-node/routes/foo/index.ts
```diff
@@ -0,0 +1,5 @@
+module.exports = async (fastify) => {
+  fastify.get("/", function () {
+    return { foo: "bar" };
+  })
+};
```
simoneb, line 5: formatting: let's use quotes consistently, either single or double

test/typescript-jest/babel-node/index.test.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  it("tests the root route", async function () {
+    const response = await app.inject({
+      method: 'GET',
+      url: '/'
+    })
```
simoneb, line 19: if you're GETting root you don't need this variant of inject, simple inject('/')

simoneb (approved):
a few comments but overall LGTM

mcollina (approved):
lgtm

mcollina:
@climba03003 PTAL

climba03003 (approved):

PR #273, opened 2022-10-12
Conditionally load tsnode for vitest

Uzlopak:
based on #264 

I think #264 was breaking, because of multiple reasons:

- it was not checking if we have a typescript file, and was probably always trying to load ts-node
- it was always trying to load ts-node even in production, which is bad

So this PR fixes that by checking if we even have a typescript file and if we are in vitest, it will load ts-node. 

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,15 +8,22 @@ const pkgUp = require('pkg-up')
 const isTsNode = (Symbol.for('ts-node.register.instance') in process) || !!process.env.TS_NODE_DEV
 const isBabelNode = (process?.execArgv || []).concat(process?.argv || []).some((arg) => arg.indexOf('babel-node') >= 0)
 
+const isVitestEnvironment = process.env.VITEST === 'true' || process.env.VITEST_WORKER_ID !== undefined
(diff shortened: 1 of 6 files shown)


climba03003 (changes requested):
I am not convinced on helping the user to registered `ts-node`.
If it should be happened, they can do the same inside their code.

Uzlopak:
I dont know. I personally dont use vitest. 

We already have code for other test frameworks regarding typescript files loading. Only vitest currently needs manually loading ts-node to work. So I am imho just offering a reasonable solution.

climba03003:
> I dont know. I personally dont use vitest.
> 
> We already have code for other test frameworks regarding typescript files loading. Only vitest currently needs manually loading ts-node to work. So I am imho just offering a reasonable solution.
> 
> btw. changed the PR description

Have you tried `forceESM` option without registered `ts-node`?

Uzlopak:
Strange! The tests pass?!

Uzlopak:
Closing in favor of #274

Issue #291, opened 2023-01-05
dirNameRoutePrefix option doesn't work when set to false


hugoattal:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

4.11.0

### Plugin version

5.7.0

### Node.js version

18.12.1

### Operating system

Windows

### Operating system version (i.e. 20.04, 11.3, 10)

Windows 11

### Description

I want to disable the `dirNameRoutePrefix` settings as my route already exposed their prefix
But routes aren't loading at all when setting `dirNameRoutePrefix: false`.

I think it comes from the fact that prefixes are used to index plugins in `hookedAccumulator`.
cf https://github.com/fastify/fastify-autoload/blob/master/index.js#L162
But I'm not 100% sure...

➡️ **If you want, I can investigate and make a PR to fix this!**

### Steps to Reproduce

Here's a small example

```ts
fastify.register(autoLoad, {
  dir: path.join(__dirname, 'plugins'),
  dirNameRoutePrefix : false
})
```

```ts
export default async (fastify: any) => {
  fastify.get("/foo/bar", function () {
    return { foo: "bar" };
  })
};
```

### Expected Behavior

Plugins/routes should load correctly.

mcollina:
Thanks for reporting!

Can you provide steps to reproduce? We often need a [reproducible example](https://stackoverflow.com/help/minimal-reproducible-example), e.g. some code that allows someone else to recreate your problem by just copying and pasting it. If it involves more than a couple of different file, create a new repository on GitHub and add a link to that.

hugoattal:
Ok, I'm really sorry, I should have dig into it a bit further... It appears that the `dirNameRoutePrefix` impacted the path sent to the `matchFilter` I had. So my `matchFilter` no longer worked.

For example, my filter function received this with the `dirNameRoutePrefix`: `/user/handlers/ping.ts`
And this without it: `///ping.ts`

100% my fault for not digging a bit. And another 100% my fault for building this very feature 😅...

I'll make a PR to fix this (else, the matchFilter and ignoreFilter don't have much use).
Again, so sorry...

PR #292, opened 2023-01-06
Fix file path for filters

hugoattal:
Fix https://github.com/fastify/fastify-autoload/issues/291

I'm so sorry, the matchFilter and ignoreFilter didn't have the correct behavior when `dirNameRoutePrefix` was altered...
So I used the real file path relative to the root dir.

⚠️ There were a problem with windows with the backslashes. I considered using [unjs/pathe](https://www.npmjs.com/package/pathe), but I felt like this was not necessary to add a package for such a tiny usecase. So I simply used a `replaceAll`, tell me if that's okay.

I also update the test routes responses to `{[name of route]: "[name of the route]"}` for better readability.
I can revert that if you prefer the old way.

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -236,7 +236,7 @@ async function findPlugins (dir, options, hookedAccumulator = {}, prefix, depth 
 
   function accumulatePlugin ({ file, type }) {
     // Replace backward slash to forward slash for consistent behavior between windows and posix.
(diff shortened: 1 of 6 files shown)


hugoattal:
Well, seems like `replaceAll` if a bit new for old node version. I'm replacing it for a regular `replace` with a RegExp.

mcollina (approved):
lgtm

PR #330, opened 2023-09-15
perf(index): convert unused capture group to non-capture group

Fdawgs:
- Changed capture group `(...)` to non-capture group `(?:...)`, as nothing is done with the captured text in this plugin, the regex engine is doing extra work capturing text only to end up not using it
- Added the `u` unicode flag to every regex pattern, which [acts as a strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/unicode#description) for regex
- Removed a redundant `i` case-insensitive flag from a pattern that had no word characters in it

#### Checklist


gurgunday (approved):

mcollina (approved):
lgtm

PR #335, opened 2023-10-18
chore: remove `getPackageType` deps, fix for Bun

ObscuritySRL:
The `getPackageType` function needlessly relied on the `pkg-up` package tree. While unnecessary, it also caused `@fastify/autoload` to not work with Bun.

Have run `lint:fix` and `test` with no issues. All changes are internal and require no documentation changes.

Not sure the other requirements for contributing to Fastify.

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -117,13 +117,13 @@ const fastifyAutoload = async function autoload (fastify, options) {
 async function getPackageType (cwd) {
   const directories = cwd.split(path.sep)
 
-  // required for paths that begin with the sep
-  directories[0] ||= path.sep
+  // required for paths that begin with the sep, such as linux root
+  directories[0] = directories[0] !== '' ? directories[0] : path.sep
 
   while (directories.length > 0) {
     const file = path.join(...directories, 'package.json')
 
-    const fileExists = await fs.access(file, fs.constants.F_OK)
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -116,9 +115,23 @@ const fastifyAutoload = async function autoload (fastify, options) {
...
+
+  while (directories.length > 0) {
+    const file = path.join(...directories, 'package.json')
+
+    const fileExists = await fs.stat(file)
+      .then(() => true)
+      .catch(() => false)
+
+    if (fileExists) {
+      return require(file).type
```
mcollina, line 131: Please use `fs.readFile` instead

index.js
```diff
@@ -1,9 +1,8 @@
 'use strict'
 
+const fs = require('node:fs/promises')
```
Fdawgs, line 3: Please destructure. :)

Fdawgs (changes requested):

index.js
ObscuritySRL: Is there any particular reason for this? I only used `require` because the previous dependencies did, but am curious as it seems fairly fool-proof.

Cheers!

mcollina (approved):
lgtm, thx

PR #280, opened 2022-11-18
add esbuild-register support

MrWook:
What I did
Add [esbuild-register](https://github.com/egoist/esbuild-register) support. One of the many libraries to execute typescript files on the fly

#### Checklist

- [x] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #346, opened 2023-12-17
feat: add typescript support for `vite` and `vite-node`

### LINKED ISSUE #226: fastify-autoload cannot import plugin...To fix this error compile TypeScript to JavaScript or use 'ts-node' to run your app.
### Prerequisites

- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

Hi,

I'm  trying to use autoload with TypeScript files but I don't understand what is wrong.

An example 

```JS
// plugins/test/module.ts
import { FastifyInstance, FastifyPluginOptions } from 'fastify'
import fp from 'fastify-plugin'

function plugin (f: FastifyInstance, opts: FastifyPluginOptions, next: () => void) {
  f.get('/', (request, reply) => {
    reply.send({ something: 'else' })
  })

  next()
}

export default fp(plugin)
```

```js
// index.js
import path from 'path'
import fastify from 'fastify'
import fastifyAutoload from 'fastify-autoload'
...
server.register(fastifyAutoload, {dir: path.join(__dirname, '../src/plugins')})
...
```

```json
// tsconfig.json
{
  "extends": "fastify-tsconfig",
  "compilerOptions": {
    "target": "es2016",
    "experimentalDecorators": true,
    "emitDecoratorMetadata": true,
    "module": "commonjs",
    "baseUrl": ".",
    "declaration": true,
    "sourceMap": true,
    "outDir": "./dist",
    "allowSyntheticDefaultImports": true,
    "esModuleInterop": true,
    "forceConsistentCasingInFileNames": true,
    "strict": true,
    "noImplicitAny": true,
    "strictNullChecks": true,
    "strictFunctionTypes": true,
    "noImplicitThis": true,
    "noImplicitReturns": true,
    "skipLibCheck": true
  },
}
```

```json
// nodemon.json
{
  "watch": ["dist"],
  "ext": "js",
  "execMap": {
    "ts": "ts-node"
  }
}
```

```json
// package.json
{
  ...
    "scripts": {
    "build": "tsc -p tsconfig.json",
    "start": "node dist/index.js",
    "watch": "tsc --watch --preserveWatchOutput --pretty",
    "dev": "cross-env NODE_ENV=development nodemon --config nodemon.json dist/index.js",
    "test": "echo \"Error: no test specified\" && exit 1"
  },
...
}
```

Any hint? 
Thank you

xeho91:
Resolves #226

#### Checklist


xeho91:
## NOTES:

- I encountered the same issue #226, and I thought there could be a way to work around it. I personally used [`vite-node`](https://github.com/vitest-dev/vitest/tree/main/packages/vite-node), but same case could happen with using [`vite`](https://github.com/vitejs/vite/tree) too.
- The solution provided is simply checking if the `basename` of the `process.argv[1]` returns a bin filename for both `vite` and `vite-node`. Stripping extension - I've left comment with link to the `package.json`, just in case they would change the extension name. Who knows.
- I am not sure if this ideal solution.
- There was no `npm run benchmark` script available, I guess this PR template is general across your repositories, so I assumed `npm run test` was enough.
- I've got no idea how to write a test/or benchmark for this case. If they're needed I would appreciate the guidance on how to test it.
- I love Fastify ❤️

xeho91:
Woops, I might have broken something!

As the tests are failing for the Node.js `v18`, however they pass for the latest version (`v21`).

I'm not knowledgeable enough to understand if this test is strictly related to the changes I implemented. As far I understood the code, and the fact that I'm only using the `basename` from standard library - `node:path`. I don't see the connection with the error message, hence why I would appreciate any insight or guidance.

xeho91:
I was able to reproduce the issue locally.

**The fix is to upgrade `tsx` to latest version - see #344.**
So, this PR depends on it in order for the test to pass.

mcollina:
I'm ok in dropping Node v14 and v16 with this PR, but it should still support Node v18.
Right now I do not have much time to dig into why this wasn't working for Vite, every time I did a solution eluded me. Can you add a test too if manage to make current one passing?

xeho91:
> I'm ok in dropping Node v14 and v16 with this PR, but it should still support Node v18. Right now I do not have much time to dig into why this wasn't working for Vite, every time I did a solution eluded me. Can you add a test too if manage to make current one passing?

1. Oh I see, upgrading `tsx` also comes with a cost of latest supported version of Node.js from `v18`.
2. Testing. Sure, I can provide tests. However, at the current moment I don't have a idea for a best test approach for it.

Once I'll have a free time, I'll sit on this issue again and will think about how to write a test for this one. If you have any hints, I'll be grateful!

> [!IMPORTANT]
> Important note as well, in case I forget again in the future.
> If using with `vite`, then we need to provide a loader:
> ```sh
> NODE_OPTIONS="--loader=tsx" vite
> ```

mcollina:
I'm ok to ship a new major and support v18.19+.

Uzlopak:
See e.g.https://github.com/fastify/fastify-autoload/pull/280/files for implementing a test

But if you can provide an example repo where you use vite and node-vite and your fork to load your vite and node-vite code, I could add the test to this PR.

xeho91:
## Update

Firstly, thank you for you patience, as the Lunar New Year is almost here, I finally have time to finish unresolved matters.

✅  So, I've managed to add test for when running with `vite-node`.

❌  However, I'm unable to configure tests properly for building via `vite`. I've had errors related to not being able to use Node.js builtin modules in the browser such as this one:

```txt
RollupError: "dirname" is not exported by "__vite-browser-external"
```

I've tried to do a lot of manipulation via `vite.config.ts`, and eventually I lost track of time trying to reproduce on how I was able to make it work during my experimenting a while ago.

---

So, if the test is really mandatory for using with `vite` _(I'm using `vite-node` instead)_... I propose that I could just drop _(remove)_ this extra attempt to support using `vite`, as I failed to provide a test for this case, but we could move forward with using `vite-node`.

mcollina:
Well, the goal was to support `vite`, not `vite-node`. I'm okay with you amending the PR only to support `vite-node`.

What's the `vite.config.ts` that you are trying to use? It seems that you are trying to bundle fastify-autoload for a browser environment, which is unlikely to work, given that it needs access to the file system.

xeho91:
> Well, the goal was to support `vite`, not `vite-node`. I'm okay with you amending the PR only to support `vite-node`.
> 
> What's the `vite.config.ts` that you are trying to use? It seems that you are trying to bundle fastify-autoload for a browser environment, which is unlikely to work, given that it needs access to the file system.

Okay, I didn't want to give up on this so easily after all.

I know the error message seems like as I was trying to build for browser, while I used the `lib` option and `target: ["node20"]`. I was clueless at this point 🤷 as to what is going on.

So, today with fresh mind, I attempted again to write tests for `vite`, and I also found on my device the experimental repository.

The missing link was that I was not using the `vite-plugin-node`.
But, this almost worked.

Sadly, I encountered this error:

```txt
RollupError: "default" is not exported by "index.js", imported by "test/typescript/basic/app.ts".
```

So, I knew I had to modify the `rollupOptions`. And it was enough just to add another plugin `@rollup/plugin-commonjs`.

And voila, it worked ✅.


**So, to summarize:**
In the provided tests for `vite`, I'm testing building for:

- `/test/typescript/`
- `/test/typescript-esm/`

In the end, I think using `vite` for bundling an Node.js app seems to be a bit too much of a hassle after all - on the setup part. Using `vite-node` for bundling the `fastify` app is just more simpler.

mcollina:
CI is failing

mcollina:
Apparently vite-plugin-node requires vite v4.0.0: https://github.com/axe-me/vite-plugin-node/blob/b9ff47f6f02b4ffd4648812a3de6d5d133879ac6/packages/vite-plugin-node/package.json#L41 

I think you can set legacy-peeer-deps in the `.npmrc`: https://github.com/npm/rfcs/discussions/283

climba03003:
Closing due to inactivity.

PR #368, opened 2024-03-08
composedPlugin should have its own prefix context

### LINKED ISSUE #326: setNotFoundHandler is not encapsulated when called inside autohooks.js
### Prerequisites

### Fastify version

4.22.2

### Plugin version

4.5.1

### Node.js version

18.7

### Operating system

Linux

20.04

### Description

Using autoload & `autohooks.js` inside a folder, when defining `setNotFoundHandler` inside `autohooks.js`, & defining `prefix` on the autoload options, the defined `setNotFoundHandler` leaks outside of the `prefix`.

### Steps to Reproduce

Assume the following folder structure.
```
├── index.ts
├── plugins
│   ├── renderer.ts
│   └── routes.ts
├── routes
│   ├── api
│   │   ├── alive-route.ts
│   │   └── autohooks.ts
│   └── renderer.ts
```

```ts
// plugins/routes.ts
export const routesPlugin: FastifyPluginCallback = async function routesPlugin(instance) {
  instance.route(rendererRoute);

  instance.register(autoload, {
    dir: path.join(__dirname, '../routes/api'),
    forceESM: true,
    autoHooks: true,
    cascadeHooks: true,
    options: { prefix: env.API_PATH }, // this is the autoload prefix its value is `/api`
  });
};
```

```ts
// routes/api/autohooks.ts
const autoHooks: FastifyPluginAsync = async function autoHooks(instance) {
  instance.setNotFoundHandler(() => {
    throw new ActionNotFound();
  });
}

export default autoHooks;
```

When calling to `/api/not-exists` it shows the `ActionNotFound` error (as expected)
but, when calling `/not-exists` it also shows the `ActionNotFound` error, evenough, the not found handler is defined inside the prefix `/api`.

I've tried even to add `encapsulate: true` to autoload options, without luck.

### Expected Behavior

From Fastify [docs](https://fastify.dev/docs/latest/Reference/Server/#setnotfoundhandler), it specifies that if there is a `prefix` defined, `setNotFoundHandler` should be encapsulated to this prefix.

AutoLoad should enforce the same thing.

jean-michelet:
Fixes #326 

Currently, when autohooks are involved, we observe unexpected behavior, such as `setNotFoundHandler` associated with the root `/` prefix even though it is defined in a child folder with its own prefix. This prevents custom configuration of `setNotFoundHandler` by child and sibling plugins.

It appears that plugins involving autohooks are composed in a parent plugin, I think that's when the prefix context is lost: https://github.com/fastify/fastify-autoload/blob/e107d5b3019c84fc193cbbb774aa52e521004c97/index.js#L103

~Not sure how `setNotFoundHandler` is associated with root prefix when defined in a child plugin but I guess it has something to do with `skip-override` configured for hookPlugins?~

In this PR, I propose to pass the prefix option to the composed plugin, but I'm not sure this is the expected behavior, wdyt?


jean-michelet:
Also fails on `master`: https://github.com/fastify/fastify-autoload/actions/runs/8201748436/job/22433147480?pr=368#step:5:770

mcollina:
Could you please merge from main? I think the tests will now pass.

mcollina (approved):
lgtm

PR #373, opened 2024-05-03
Update for V5 (bis)

jean-michelet:
Read  #371 for context.

Some commits have not been merged on `next`, shouldn't they?


package.json
```diff
@@ -1,16 +1,17 @@
...
   "main": "index.js",
   "type": "commonjs",
   "types": "types/index.d.ts",
   "scripts": {
     "lint": "standard | snazzy",
     "lint:fix": "standard --fix | snazzy",
-    "test": "npm run unit && npm run typescript && npm run typescript:jest && npm run typescript:esm && npm run typescript:swc && npm run typescript:swc-node-register && npm run typescript:tsm && npm run typescript:tsx && npm run typescript:vitest && npm run typescript:esbuild",
-    "typescript": "tsd",
+    "test": "npm run typescript && npm run typescript:jest && npm run typescript:default && npm run typescript:swc && npm run typescript:swc-node-register && npm run typescript:tsm && npm run typescript:tsx && npm run typescript:vitest && npm run typescript:esbuild && npm run unit",
+    "typescript": "tap plugin add @tapjs/typescript && tsd",
```
jean-michelet, line 12: I'm really dissatisfied with launching this type of command, but I don't know how to do better with Tap 18.

package.json
```diff
@@ -19,8 +20,8 @@
     "typescript:esbuild": "node scripts/unit-typescript-esbuild.js",
     "typescript:vitest": "vitest run",
     "typescript:vitest:dev": "vitest",
-    "unit": "node scripts/unit.js",
-    "unit:with-modules": "tap test/issues/*/test.js test/commonjs/*.js test/module/*.js test/typescript/*.ts"
+    "unit": "tap plugin rm @tapjs/typescript && node scripts/unit.js",
```
jean-michelet, line 23: The typescripts files and other scripts are launched first and tests finish with `node scripts/unit.js`.

scripts/unit-typescript.js
```diff
@@ -0,0 +1,17 @@
+'use strict'
+
+const { exec } = require('node:child_process')
+
+const args = [
+  'tap',
+  '--disable-coverage',
+  'test/typescript/*.ts'
```
jean-michelet, line 8: We can move this test here, the coverage check is not affected.

test/commonjs/autohooks/routes/.autohooks.js
```diff
@@ -1,7 +1,6 @@
-module.exports = async function (app, opts, next) {
```
jean-michelet, line 1: There were tons of warning indicating deprecation for async/callback mix but CI currently don't fail.

.taprc
```diff
@@ -1,3 +1,3 @@
-ts: true
-jsx: false
-flow: false
+# vim: set filetype=yaml :
+plugin:
+  - "!@tapjs/typescript"
```
jean-michelet, line 3: This is automatically added by tap, I can commit the empty file.

jean-michelet:
I experimented these timeout fails several times...

gurgunday:
Interestingly ubuntu is failing and macos is not

package.json
jean-michelet: I'm considering exploring alternative approaches, I really don't like the "fix" I am doing. I am gonna delve deeper into their documentation and source code  and eventually open an issue on their repository. I could be missing something as I don't really know tap.

@mcollina @climba03003, do you have any suggestions for me?

jean-michelet:
I am gonna look for these timeouts this afternoon, CI fail 50% of the time or something like that.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@fastify/autoload",
-  "version": "5.8.1",
+  "version": "5.8.2",
```
mcollina, line 3: please revert this

gurgunday:
@climba03003 can you chime in? do you have an idea why this might be happening?

jean-michelet:
Thanks for the sync.

As I said to @gurgunday in private, it seems to fail on Ubuntu 18 specifically:
- https://github.com/jean-michelet/fastify-autoload/actions/runs/8938089684/job/24551748911
- https://github.com/jean-michelet/fastify-autoload/actions/runs/8938790498/job/24553609139
- https://github.com/jean-michelet/fastify-autoload/actions/runs/8948405642/job/24581484388

jean-michelet:
Hmmm, no there's a failure on windows too, I hadn't noticed that.

.taprc
```diff
@@ -0,0 +1,3 @@
+# vim: set filetype=yaml :
+timeout: 60
+plugin: []
```
jean-michelet, line 3: `.taprc` is updated each time tests are run.
It adds the vim modeline and the set plugin to `[]`.
When validating `git commit -a -m`, considering the githook precommit that runs the tests, this is the final result.
I guess it's more to to keep it that way?

package.json
climba03003: I don't think we should rely on `@tapjs/typescript`
It means when we are running under `tap`, it is trans-pile by `tap` instead of the loader.

package.json
```diff
@@ -47,12 +45,13 @@
   "devDependencies": {
     "@fastify/pre-commit": "^2.1.0",
     "@fastify/url-data": "^5.4.0",
-    "@swc-node/register": "^1.9.0",
+    "@swc-node/register": "^1.9.1",
     "@swc/core": "^1.4.13",
-    "@swc/register": "^0.1.9",
```
jean-michelet, line 52: > [This package has been deprecated](https://www.npmjs.com/package/@swc/register)

.swcrc
jean-michelet, line 1: Do we really need this config?

Btw, we must pass env variable `SWCRC=true` to load it.

jean-michelet (commented):
I removed the use of tap in `unit-typescript-*` and run them directly with node.
I think it's ok, do I miss something?

jean-michelet:
I've done what I can to help @gurgunday , but I'm getting a bit tired with this PR. I don't have enough experience or time currently, I think.

If someone more competent than me wants to finish it :sweat_smile:

PR #378, opened 2024-05-29
fix: autohooks prefix concatenation

### LINKED ISSUE #376: Fastify 5.8.2 and autohooks produces "route already declared" error
### Prerequisites

### Fastify version

4.27.0

### Plugin version

5.8.2

### Node.js version

20.13.1

### Operating system

macOS

14.5

### Description

When going from 5.8.0 to 5.8.2 my code base will throw a `FastifyError [Error]: Method 'GET' already declared for route '/backend'` error about having a duplicate route. I managed to reproduce this and added a simple repository to show the error.

https://github.com/mccare/fastify-autoload-hook

### Expected Behavior

It should just register the route into /backend/example (in this case) and not try to register under /backend

climba03003:
Since no core member add value to https://github.com/fastify/fastify-autoload/issues/376#issuecomment-2129429693
I want to move things forward to users desire behavior.

Fixes #376 

As mention in last submitted PR, I believe the long term solution is refactor the code to create a single tree that contain both `hook` and `plugin` information.

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -26,6 +26,11 @@ const fastifyAutoload = async function autoload (fastify, options) {
   const pluginArray = [].concat.apply([], Object.values(pluginTree).map(o => o.plugins))
   const hookArray = [].concat.apply([], Object.values(pluginTree).map(o => o.hooks))
 
+  let rootPrefix = options.options?.prefix ?? ''
+  // when prefix is provided /prefix/ format
+  // it is not good for concat, since most folder prefix start with /
+  if (rootPrefix[rootPrefix.length - 1] === '/') rootPrefix = rootPrefix.slice(0, -1)
+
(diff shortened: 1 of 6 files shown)


gurgunday (approved):
I didn't provide feedback before as I had no strong opinions and have never used the library

If you think this is better, I am behind it

gurgunday (approved):
Yeah this looks good, I was going to merge, but this is breaking right?

Should we merge to `next`?

climba03003:
Changing of test seems to indicate this is breaking, but actually not.

The new behavior is introduced in 5.8.1 and it was never tested before.
And that is the one introducing new breaking behavior (or called never clarified behavior).

Here is just reverting to behavior to the user most expected one.

mcollina (approved):
lgtm

Thanks @climba03003, I've been buried down in work.

PR #381, opened 2024-06-11
Allow `autoConfig` with callback

jean-michelet:
Can be usefull to autoload plugins:
```js
import fastifyMysql from "@fastify/mysql";

export const autoConfig = (fastify) => {
  return {
    promise: true,
    host: fastify.config.MYSQL_HOST,
    user: fastify.config.MYSQL_USER,
    password: fastify.config.MYSQL_PASSWORD,
    database: fastify.config.MYSQL_DATABASE,
    port: Number(fastify.config.MYSQL_PORT)
  }
}

export default fastifyMysql
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -308,7 +308,7 @@ FASTIFY_AUTOLOAD_TYPESCRIPT=1 node --loader=my-custom-loader index.ts
 
 Each plugin can be individually configured using the following module properties:
 
-- `plugin.autoConfig` - Configuration object which will be used as `opts` parameter
+- `plugin.autoConfig` - Specifies the options to be used as the `opts` parameter.
 
   ```js
   module.exports = function (fastify, opts, next) {
@@ -330,6 +330,13 @@ Each plugin can be individually configured using the following module properties
     })
   }
   export const autoConfig = { name: 'y' }
+  ```
+
+  You can also use a callback function if you need to access the parent instance:
+  ```js
+  export const autoConfig = (fastify) => {
+    return { name: 'y ' + fastify.rootName }
+  }
   ```
 
 - `plugin.autoPrefix` - Set routing prefix for plugin
--- a/README.md
+++ b/README.md
@@ -339,6 +339,15 @@ Each plugin can be individually configured using the following module properties
   }
   ```
 
+  However, note that the `prefix` option should be set directly on `autoConfig` for autoloading to work as expected:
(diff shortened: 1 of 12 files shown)


README.md
```diff
@@ -332,6 +332,22 @@ Each plugin can be individually configured using the following module properties
   export const autoConfig = { name: 'y' }
   ```
 
+  You can also use a callback function if you need to access the parent instance:
+  ```js
+  export const autoConfig = (fastify) => {
```
mcollina, line 337: I think we should pass in the `overrideConfig` that was previously merged in

mcollina (changes requested):

README.md
jean-michelet: Oops... Sorry...



Would it be ok?

```js
let pluginOptions
  if (typeof pluginConfig === 'function') {
    pluginOptions = function (fastify) {
      return { ...pluginConfig(fastify), ...overrideConfig};
    }
  } else {
    pluginOptions = { ...pluginConfig, ...overrideConfig}
  }
 ```

Comment (jean-michelet) [reply README.md:337]: Tests are not broken by this code, but I didn't test to ensure it works for now.

File: index.js [hunk 2]
```diff
@@ -282,7 +282,17 @@ async function loadPlugin ({ file, type, directoryPrefix, options, log }) {
...
+  if (typeof pluginConfig === 'function') {
+    pluginOptions = function (fastify) {
+      return { ...pluginConfig(fastify), ...overrideConfig }
+    }
+
+    pluginOptions.prefix = pluginConfig.prefix
+  } else {
+    pluginOptions = { ...pluginConfig, ...overrideConfig }
+  }
+
```
jean-michelet, line 295: That might do the job, but I think the whole component would benefit greatly from a major refactoring.
Right now, it's really looks spaghetti code imho, it's hard to introduce changes cleanly.

jean-michelet: Open to work on it btw, but after the demo.

mcollina: That'd be fantastic, thx 'b

mcollina (approved):
lgtm

mcollina:
@climba03003 ptal

climba03003 (approved):
LGTM

index.js
jean-michelet: See #383

PR #396, opened 2024-07-13
Merge `next` into `master`

jsumners:
As titled.

PR #398, opened 2024-07-16
Increase jest timeout

jean-michelet:
Related to #396

Not sure how to fix this issue, I try to increase the timeout to see if the CI pass.


jean-michelet:
Ok, the test is staling...

Issue #383, opened 2024-06-17
Refactoring the component


jean-michelet:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the issue has not already been raised

### Issue

The code is becoming increasingly difficult to maintain, I think it deserves a redesign as suggested [here](https://github.com/fastify/fastify-autoload/issues/380#issuecomment-2159851069) and [here](https://github.com/fastify/fastify-autoload/pull/381#discussion_r1635303218).

I'd like to know if you have any suggestions before proceeding, we could also consider proceeding step by step.

Uzlopak:
I dont see the maintenance problem. The code is actually of very high quality. Also it is "only" about 500 LOC. The issue is that the problem it tries to solve is complex itself. That results in not seeing every "edge" case.

jean-michelet:
Perhaps I should have expressed myself less assertively.
I have the feeling that the code could be better organized and more readable, but this not the reason I want to invest time in it. 
Would refactoring help solve the issue problem and eventually add new features? If yes, how?

Maybe you can elaborate what you were thinking by saying this @climba03003:
> Rework the files finding and tree building - it is the critical one because all the information is scattered currently.

climba03003:
The main problem is that the current information split into two array.
https://github.com/fastify/fastify-autoload/blob/13706979a74cfac4c414d9114480c04f2c52833d/index.js#L26-L27
Sometimes the information is not about to find inside some function.

For what I propose is that we should define the works into 3 clear steps.
1. Finds all available files including the config.
2. Build the tree based on step 1. and the plugin configure
3. Registered plugin based on the tree in step 2.

In this case we can always change the function inside step 2. to provide different behavior instead of looking around different function and gather necessary information in different place.

jean-michelet:
I agree that hooks should be composed to plugins before registration step.

> Build the tree based on step 1. and the plugin config

Btw, do you think we should allow `auto` exports from autohooks files too?
```ts
export const autoConfig = { name: 'y' }
export const autoPrefix = "/prefixed";
```

I am not sure that it's a good idea but some people want to use them there, I saw this in a [reproduction issue repo](https://github.com/rommelandrea/fastify-autoload-bug/blob/a2720e87a4c5d6fbfc6029868989774e525d1da8/routes/example/autohooks.js#L7).

If we don't maybe, we should throw a clean error to educate users about it.

climba03003:
[...]

PR #401, opened 2024-07-26
feat: reduce cognitive complexity

jean-michelet:
This PR is intended to reduce the overall cognitive complexity of the component, which is imho unnecessarily hard to read, although I understand this is subjective. Basically, I've done a bit of refactoring, breaking down huge functions into smaller ones and limiting the use of nested closures.

I put the `findPlugins` function in its own file to reduce the size of `index.js` and because it is a clearly separated step.

I tried to do the same with `loadPlugins` but `vitest` failed to transpile after that... But the two big steps are finding and loading, registering is just a few lines so I think it's ok.

I think I've also slightly reduced algorithmic complexity by making a single global iteration on `pluginTree`, including plugin loading, hooks and registration. This also makes it possible to benefit from the current iteration context, instead of making distinct tree traversal for hooks and plugins, although this point isn't quite clear to me yet. So it's not a fix of #383.


find-plugins.js
```diff
@@ -0,0 +1,181 @@
+const { readdir } = require('node:fs/promises')
```
mcollina, line 1:
```suggestion
'use strict'

const { readdir } = require('node:fs/promises')
```

find-plugins.js
```diff
@@ -0,0 +1,181 @@
...
+const { relative, join } = require('path')
+const runtime = require('./runtime')
```
mcollina, line 3: can you move this file and `runtime` into `lib/`?

find-plugins.js
```diff
@@ -0,0 +1,181 @@
...
+    if (dirent.isDirectory() && !atMaxDepth) {
+      processDirectory({ prefix, opts, dirent, dir, file, directoryPromises, hookedAccumulator, depth, currentHooks })
+    } else if (indexDirent) {
+      // An index.js file is present in the directory so we ignore the others modules (but not the subdirectories)
+    } else if (dirent.isFile() && opts.scriptPattern.test(dirent.name)) {
+      processFile({ file, opts, dirent, hookedAccumulator, prefix })
+    }
+  }
+
+  await Promise.all(directoryPromises)
```
mcollina, line 95: Can you avoid having a `directoryPromises` array at all? I think this is one of the things that makes this code very hard to read. We should be making `processDirectory` `async` instead and `await` it in place.

jean-michelet: Yes.

mcollina (approved):
lgtm

jean-michelet:
Just look for my last commit plz, I had missed a previous optimization during refactorisation.

gurgunday (approved):
LGTM

gurgunday:
Can you update the workflow to version 5?

https://github.com/fastify/fastify-autoload/blob/3584972aa271aed85c4443b759f8c53a640db773/.github/workflows/ci.yml#L20

jean-michelet:
I do it in an independent PR @gurgunday : https://github.com/fastify/fastify-autoload/pull/403

jean-michelet:
I would prefer a review from @climba03003 before merging.

jean-michelet:
Can you review @climba03003 plz or can we merge?

climba03003 (approved):
LGTM.

PR #406, opened 2024-07-31
fix: check require.cache to find preloaded modules

### LINKED ISSUE #404: Cannot detect `@swc-node/register` loaded by `fastify-cli` with the `-r` option
### Prerequisites

### Fastify version

4.28.1

### Plugin version

5.10.0

### Node.js version

20.x

### Operating system

Windows

10

### Description

I'm using `fastify start -r @swc-node/register app.ts` to start my app written in TypeScript.

However, this plugin cannot detect `@swc-node/register` loaded by `fastify-cli` and the following error is thrown:

```console
Error: @fastify/autoload cannot import plugin at '~/fastify-app/plugins/sensible.ts'. To fix this error compile TypeScript to JavaScript or use 
'ts-node' to run your app.
```

**Related PR:**

PR https://github.com/fastify/fastify-autoload/pull/191 fixed the case with `node -r @swc-node/register` but not with `fastify start -r @swc-node/register`.

**Current workaround:**

I have to manually set the `FASTIFY_AUTOLOAD_TYPESCRIPT` environment variable.

**Additional information:**

The current `checkPreloadModules()` util checks the `process._preload_modules` array, but it's empty when the package is required with `fastify-cli`.

https://github.com/zetaraku/fastify-app

### Expected Behavior

The plugin should be able to detect the `@swc-node/register` package loaded by `fastify start -r @swc-node/register`.

jean-michelet:
Fixes #404 

I tested on the reproduction repo and it does work well: https://github.com/zetaraku/fastify-app

### DIFF
--- a/runtime.js
+++ b/runtime.js
@@ -1,5 +1,7 @@
 'use strict'
 
+const path = require('path')
+
 // runtime cache
 const cache = {}
 
@@ -14,11 +16,16 @@ function checkProcessArgv (moduleName) {
 
 let preloadModules
 function checkPreloadModules (moduleName) {
+  const modulePath = path.join(process.cwd(), 'node_modules/', moduleName)
+
   /* c8 ignore start */
-  // nullish needed for non Node.js runtime
+  // coverage - nullish needed for non Node.js runtime
   preloadModules ??= (process._preload_modules ?? [])
+
+  // coverage - TS specific
+  return preloadModules.includes(moduleName) ||
+    Object.keys(require.cache).some(k => k.startsWith(modulePath) && preloadModules.push(modulePath))
   /* c8 ignore stop */
-  return preloadModules.includes(moduleName)
 }
 
 let preloadModulesString
--- a/runtime.js
+++ b/runtime.js
@@ -16,8 +16,6 @@ function checkProcessArgv (moduleName) {
 
 let preloadModules
 function checkPreloadModules (moduleName) {
-  const modulePath = path.join(process.cwd(), 'node_modules/', moduleName)
-
   /* c8 ignore start */
   // coverage - nullish needed for non Node.js runtime
   preloadModules ??= (process._preload_modules ?? [])
@@ -27,6 +25,7 @@ function checkPreloadModules (moduleName) {
     return true
   }
 
+  const modulePath = path.join(process.cwd(), 'node_modules/', moduleName)
   if (Object.keys(require.cache).some(k => k.startsWith(modulePath))) {
     preloadModules.push(moduleName)
 
--- a/runtime.js
+++ b/runtime.js
@@ -23,9 +23,18 @@ function checkPreloadModules (moduleName) {
   preloadModules ??= (process._preload_modules ?? [])
 
   // coverage - TS specific
-  return preloadModules.includes(moduleName) ||
-    Object.keys(require.cache).some(k => k.startsWith(modulePath) && preloadModules.push(modulePath))
+  if (preloadModules.includes(moduleName)) {
+    return true
+  }
+
+  if (Object.keys(require.cache).some(k => k.startsWith(modulePath))) {
+    preloadModules.push(moduleName)
+
+    return true
+  }
   /* c8 ignore stop */
+
+  return false
 }
 
 let preloadModulesString
--- a/runtime.js
+++ b/runtime.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const path = require('path')
+const path = require('node:path')
 
 // runtime cache
 const cache = {}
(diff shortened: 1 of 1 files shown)


runtime.js
```diff
@@ -14,11 +16,16 @@ function checkProcessArgv (moduleName) {
 
 let preloadModules
 function checkPreloadModules (moduleName) {
+  const modulePath = path.join(process.cwd(), 'node_modules/', moduleName)
```
jean-michelet, line 19: Not sure about this path join.
I just want to do a `startsWith` instead of `includes`.

runtime.js
```diff
@@ -14,11 +16,16 @@ function checkProcessArgv (moduleName) {
...
+
   /* c8 ignore start */
-  // nullish needed for non Node.js runtime
+  // coverage - nullish needed for non Node.js runtime
   preloadModules ??= (process._preload_modules ?? [])
+
+  // coverage - TS specific
+  return preloadModules.includes(moduleName) ||
+    Object.keys(require.cache).some(k => k.startsWith(modulePath) && preloadModules.push(modulePath))
```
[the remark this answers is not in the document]
zetaraku: It's not a good idea to mutate `preloadModules` in `Array.some()` and relying the truthy return value of `Array.push()`.
By the way, pushing `modulePath` instead of `moduleName` here is kinda incorrect.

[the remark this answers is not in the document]
jean-michelet: It's weird because I overthought this to avoid a couple of lines and didn't found an obvious issue.

If `preloadModules.includes(moduleName)` evaluate to true, LogicalOr right expression is not evaluated.
`some` stop iterating if evaluate truthy, LogicalAnd right is always evaluated and `push` always return truthy value when adding an element.

I refactored because I think you're right, the code is misleading and can cause problems.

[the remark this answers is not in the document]
jean-michelet: >  By the way, pushing modulePath instead of moduleName here is kinda incorrect.

Fixed, thanks.

runtime.js
```diff
@@ -15,10 +17,23 @@ function checkProcessArgv (moduleName) {
...
   preloadModules ??= (process._preload_modules ?? [])
+
+  // coverage - TS specific
+  if (preloadModules.includes(moduleName)) {
+    return true
+  }
+
+  const modulePath = path.join(process.cwd(), 'node_modules', moduleName)
+  if (Object.keys(require.cache).some(k => k.startsWith(modulePath))) {
+    preloadModules.push(moduleName)
```
zetaraku, line 30: I still discourage mutating `preloadModules`, as you may mutate the `process._preload_modules`.

runtime.js
```diff
@@ -1,5 +1,7 @@
 'use strict'
 
+const path = require('path')
```
zetaraku, line 3: For convention and best practice, use `require('node:path')` instead.

runtime.js
```diff
@@ -15,10 +17,23 @@ function checkProcessArgv (moduleName) {
...
+  // coverage - nullish needed for non Node.js runtime
   preloadModules ??= (process._preload_modules ?? [])
+
+  // coverage - TS specific
+  if (preloadModules.includes(moduleName)) {
+    return true
+  }
+
+  const modulePath = path.join(process.cwd(), 'node_modules', moduleName)
+  if (Object.keys(require.cache).some(k => k.startsWith(modulePath))) {
```
zetaraku, line 29: The convention of this repo is using parentheses even with only one parameter, and `k` can be something more readable name IMO.

runtime.js
zetaraku: Can `process.cwd()` guarantee to be the project root if executed from other path?

runtime.js
```diff
@@ -15,10 +17,23 @@ function checkProcessArgv (moduleName) {
...
+  // coverage - nullish needed for non Node.js runtime
+  preloadModules ??= [...(process._preload_modules ?? [])]
+
+  // coverage - TS specific
+  if (preloadModules.includes(moduleName)) {
+    return true
+  }
+
+  const modulePath = path.join(process.cwd(), 'node_modules', moduleName)
+  if (Object.keys(require.cache).some((k) => k.startsWith(modulePath))) {
```
climba03003, line 29: What if we run under `ESM`?

climba03003 (dismissed):

runtime.js
jean-michelet: https://github.com/fastify/fastify-autoload/pull/406/files#diff-620d8f84607e4a3c91579ab868c89080849559f6533315da0c0ec092f9899a11R21

runtime.js
jean-michelet: What would you do?

runtime.js
jean-michelet: I think I am gonna do a PR specific for immutability as there is a lot of chance that this PR will close. I don't feel confident with it.

mcollina:
Thanks for opening a PR! Can you please add a unit test?

jean-michelet:
@zetaraku seems to prefer this to be fixed in fastify-cli.
Maybe we should investigate, plus I am not sure how to write relevant unit tests for this.

Are you willing to push a PR on cli @zetaraku? As a POC?

zetaraku:
Will try to make one later.

PR #408, opened 2024-08-24
feat: Add Bun and Deno to typescript support check

### LINKED ISSUE #317: Fastify Autoload does not work with Bun.js or deno
### Prerequisites

### Fastify version

4.7.0

### Plugin version

5.7.0

### Node.js version

18.16.1

### Operating system

Linux

38

### Description

When using bun to run fastify with the fastify autoload plugin, the code crashes with the following error:

```bash
124 |       return join(__dirname, '..', 'file.js')
125 |     }
126 | 
127 |     let fixTarget
128 | 
129 |     for (const filePath of callers) {
                               ^
TypeError: undefined is not an object (evaluating 'callers')
      at fixTarget (/home/notsql/Development/Flora/back-end/node_modules/pino/lib/transport.js:129:27)
      at transport (/home/notsql/Development/Flora/back-end/node_modules/pino/lib/transport.js:114:21)
      at normalizeArgs (/home/notsql/Development/Flora/back-end/node_modules/pino/lib/tools.js:311:15)
      at pino (/home/notsql/Development/Flora/back-end/node_modules/pino/pino.js:87:27)
      at createPinoLogger (/home/notsql/Development/Flora/back-end/node_modules/fastify/lib/logger.js:42:13)
      at createLogger (/home/notsql/Development/Flora/back-end/node_modules/fastify/lib/logger.js:101:17)
      at fastify (/home/notsql/Development/Flora/back-end/node_modules/fastify/fastify.js:137:32)
      at build (/home/notsql/Development/Flora/back-end/src/main.ts:7:14)
      at /home/notsql/Development/Flora/back-end/src/server.ts:7:15
```

### Steps to Reproduce

1. Setup fastify with fastify autoloader plugin.

2. Ensure all dependencies are installed
```bash
bun install
```

3. Run the code
```bash
bun run index.ts
```

### Expected Behavior

_No response_

Codex-:
This change allows the Bun and Deno runtimes to use this loader with Fastify base without erroring as seen in #317.

#### Node

No regressions, everything passing.

#### Bun

- `bun run typescript` ✅
- `bun run typescript:jest` ✅
- `bun run typescript:vitest` ✅
- `bun run unit` ✅
  - `unit.js` updated to consume an arg for the runtime
- The bun equivalent of the loader tests that execute `test/typescript/basic.ts` cannot be run as a particular function Tap uses has not yet been implemented in Bun
  - `NotImplementedError: node:v8 Serializer is not yet implemented in Bun.`

#### Deno

- `deno task typescript` ✅
- Deno implements `ServerResponse` as a class, while node a function, which causes this error in Fastify.
  - `Class constructor ServerResponse cannot be invoked without 'new'`
  - This causes the following to fail
    - `deno task typescript:jest` ❌
    - `deno task typescript:vitest` ❌
    - `deno task unit` ❌

While it has already been communicated that Fastify is not actively supporting either Bun or Deno runtimes, despite the results, it would be great to see this check updated to allow their usage at a "ymmv" degree.

Fixes #317 

#### Checklist

- [x] run `npm run test`
- [x] tests are included
- [x] ~documentation is changed or added~

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,3 +21,28 @@ jobs:
     with:
       license-check: true
       lint: true
+  bun:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup Bun
+        uses: oven-sh/setup-bun@v2
+
+      - name: Install dependencies
+        # The erraneous typod version `5.0.0-aplha` takes priority over .3 as 'p' > 'l'
+        run: |
+          bun i --ignore-scripts
+          bun i -D --ignore-scripts fastify@5.0.0-alpha.3
+
+      - name: Run types check
+        run: bun run typescript
+
+      - name: Run Jest tests
+        run: bun run typescript:jest
+
+      - name: Run Vitest tests
+        run: bun run typescript:vitest
+
+      - name: Run unit tests
+        run: bun run unit:bun
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,8 +23,11 @@ jobs:
       lint: true
   bun:
     runs-on: ubuntu-latest
+    continue-on-error: true
     steps:
       - uses: actions/checkout@v4
+        with:
+          persist-credentials: false
 
       - name: Setup Bun
         uses: oven-sh/setup-bun@v2
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,7 @@ jobs:
         # The erraneous typod version `5.0.0-aplha` takes priority over .3 as 'p' > 'l'
         run: |
           bun i --ignore-scripts
-          bun i -D --ignore-scripts fastify@5.0.0-alpha.3
+          bun i -D --ignore-scripts fastify@5.0.0-alpha.4
 
       - name: Run types check
         run: bun run typescript
--- a/scripts/unit.js
+++ b/scripts/unit.js
@@ -3,7 +3,7 @@
 const { exec } = require('node:child_process')
 const { argv } = require('node:process')
 
-const child = exec(argv[2] + ' run unit:with-modules')
+const child = exec((argv[2] ?? 'npm') + ' run unit:with-modules')
 
 child.stdout.pipe(process.stdout)
 child.stderr.pipe(process.stderr)
--- a/scripts/unit.js
+++ b/scripts/unit.js
@@ -1,8 +1,9 @@
 'use strict'
 
 const { exec } = require('node:child_process')
+const { argv } = require('node:process')
 
-const child = exec('npm run unit:with-modules')
+const child = exec(argv[2] + ' run unit:with-modules')
 
 child.stdout.pipe(process.stdout)
 child.stderr.pipe(process.stderr)
--- a/lib/find-plugins.js
+++ b/lib/find-plugins.js
@@ -145,7 +145,10 @@ function accumulatePlugin ({ file, type, opts, pluginTree, prefix }) {
 
 function handleTypeScriptSupport (file, language, isHook = false) {
   if (language === 'typescript' && !runtime.supportTypeScript) {
-    throw new Error(`@fastify/autoload cannot import ${isHook ? 'hooks ' : ''}plugin at '${file}'. To fix this error compile TypeScript to JavaScript or use 'ts-node' to run your app.`)
+    throw new Error(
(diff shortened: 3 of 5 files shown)


mcollina (changes requested):
Currently we do not support Bun or Deno on Fastify.
Frankly, we have not time, and I would not accept a PR without tests either.

https://github.com/fastify/fastify/blob/main/docs/Reference/LTS.md#long-term-support

Codex-:
> Currently we do not support Bun or Deno on Fastify. 

While I understand that is the official stance, as noted in your link, with all of the other supported tooling already included would it really do any harm to simply allow this check to pass for these runtimes? 

>and I would not accept a PR without tests either.

You beat me to it, I amended my commit with the bun tests I intended to push originally :)

Fastify is simply incompatible with Deno currently, I'm happy to include another test run file to run against deno but it will fail and not add value

scripts/unit.js
```diff
@@ -1,8 +1,9 @@
 'use strict'
 
 const { exec } = require('node:child_process')
+const { argv } = require('node:process')
 
-const child = exec('npm run unit:with-modules')
+const child = exec(argv[2] + ' run unit:with-modules')
```
mcollina, line 6: Can you please add `npm` as a default?

You'd need to add `bun` to CI as well.

Codex-: Sure can

I was going to ask actually, so the CI currently uses that [shared workflow](https://github.com/fastify/workflows/blob/v5.0.0/.github/workflows/plugins-ci.yml) from the workflows repo, so to not muddy concerns in that workflow, would you be happy with an additional job added to the `ci.yml` that performs the tests for bun?

```yml
jobs:
  test:
    uses: fastify/workflows/.github/workflows/plugins-ci.yml@v5.0.0
    with:
      license-check: true
      lint: true
  bun:
    uses: ...
```

Codex-: Went with adding a job to the existing workflow for now, the erroneously published version `aplha` takes precedence over `alpha` in string ordering so until it reaches beta the workflow needs to specifically install the version required for tests to satisfy the `semver` usage (since `semver` doesn't recognise `aplha` as a valid version)

.github/workflows/ci.yml
```diff
@@ -21,3 +21,28 @@ jobs:
     with:
       license-check: true
       lint: true
+  bun:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
```
Eomm, line 27:
```suggestion
      - uses: actions/checkout@v4
        with:
          persist-credentials: false
```

.github/workflows/ci.yml
```diff
@@ -21,3 +21,28 @@ jobs:
     with:
       license-check: true
       lint: true
+  bun:
+    runs-on: ubuntu-latest
```
Eomm, line 25:
```suggestion
    runs-on: ubuntu-latest
    continue-on-error: true
```

mcollina (approved):
lgtm

.github/workflows/ci.yml
```diff
@@ -21,3 +21,31 @@ jobs:
...
+          persist-credentials: false
+
+      - name: Setup Bun
+        uses: oven-sh/setup-bun@v2
+
+      - name: Install dependencies
+        # The erraneous typod version `5.0.0-aplha` takes priority over .3 as 'p' > 'l'
+        run: |
+          bun i --ignore-scripts
+          bun i -D --ignore-scripts fastify@5.0.0-alpha.3
```
gurgunday, line 39: This is not ideal

The typod version seems to be unpublished now — can't find it at https://www.npmjs.com/package/fastify?activeTab=versions

Can you try removing the line?

```suggestion
```

Codex-: You can't unpublish a package from npm:

image

gurgunday (approved):

.github/workflows/ci.yml
Codex-: Because of the typo in the version name bun takes `p` as being of a higher version than `l` when parsing `aplha`

`aplha.1` does not pass the tests

Can be removed once beta has begun

MrEAlderson:
Please merge, this is the only problem I have faced regarding bun support in fastify.
For anyone else stumbling on this: Adding `FASTIFY_AUTOLOAD_TYPESCRIPT=true` to your .env is a temporary fix

.github/workflows/ci.yml
gurgunday: I can confirm this is the case

Uzlopak:
Superseeded by #480

PR #415, opened 2024-10-08
Fix dependencies across directories

mcollina:
Fixes a regression introduced in https://github.com/fastify/fastify-autoload/pull/401 where `dependencies` across different prefixes were not taken into account.


jean-michelet (commented):
Looks good, but is it possible to write a test that proves the bug is fixed? And to avoid same regression in the future?

jean-michelet:
If too cumbersome to write, but you're sure the bug is fixed, I recommend to merge and create an issue to write this test and you assign to me.

mcollina:
> Looks good, but is it possible to write a test that proves the bug is fixed? And to avoid same regression in the future?

It's covered by moving the file `plugin-d.js` as you can see in the diff.

jean-michelet:
Ok, missed it.

jean-michelet (approved):

PR #422, opened 2024-12-01
chore(package): add funding and contribs

Fdawgs:
#### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,11 +38,28 @@
     "auto"
   ],
   "author": "Matteo Collina <hello@matteocollina.com>",
+  "contributors": [
+    {
+      "name": "Frazer Smith",
+      "email": "frazer.dev@icloud.com",
+      "url": "https://github.com/fdawgs"
+    }
+  ],
   "license": "MIT",
   "bugs": {
     "url": "https://github.com/fastify/fastify-autoload/issues"
   },
   "homepage": "https://github.com/fastify/fastify-autoload#readme",
+  "funding": [
+    {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -38,11 +38,28 @@
     "auto"
   ],
   "author": "Matteo Collina <hello@matteocollina.com>",
+  "contributors": [
+    {
+      "name": "Frazer Smith",
```
Eomm, line 43: TBH I think all or nothing here

Fdawgs: Is a bit of a pain to go through every repo and get all the contributors so was hoping everyone would add themselves.

Eomm: > Is a bit of a pain to go through every repo and get all the contributors so was hoping everyone would add themselves.

let's keep valid this one, but I would like to check in the npm page what is happning.

Right now, without the `contributors` array, the output shows the npm org team:

image

Not sure if now it will be overwritten

Fdawgs: It won't be overwritten, the collaborator list on the npm page is users who have publish privileges. If you look at the [main Fastify package](https://www.npmjs.com/package/fastify) it only shows 4 collaborators even when the collaborators list in the package.json has 20 people.

Fdawgs: It won't be overwritten, the collaborator list on the npm page is users who have publish privileges. If you look at the [main Fastify package](https://www.npmjs.com/package/fastify) it only shows 4 collaborators even when the collaborators list in the package.json has 20 people.

Issue #441, opened 2025-01-23
support native Node.js Typescript type-stripping


mcollina:
Given that Node.js v23 has unflagged type stripping, I think we should support it out of the box

mcollina:
cc @SamSalvatico @jean-michelet

jean-michelet:
We should wait for #440 to be merged.
Do you want to do it @SamSalvatico, or I take this one?

SamSalvatico:
I can take It @jean-michelet

SamSalvatico:
@jean-michelet I gave a look, but unfortunately I am not able to progress into this (missing knowledge on my side)

jean-michelet:
Have you had a look at [runtime module](https://github.com/fastify/fastify-autoload/blob/master/lib/runtime.js)?

I think we should detect the use [here](https://github.com/fastify/fastify-autoload/blob/69d42a56bfbbf9f2efd9ae01b8ddbfbfdd3a4ac9/lib/runtime.js#L107). But I haven't tested this new Node feature yet, I'll have a look.

SamSalvatico:
> runtime module

I gave a look to those links, but I cannot find a way in which the auto enable type stripping is checked, because it only looks for `checkEnvVariable('FASTIFY_AUTOLOAD_TYPESCRIPT')` or external tools, not node version

jean-michelet:
> it only looks for checkEnvVariable('FASTIFY_AUTOLOAD_TYPESCRIPT') or external tools, not node version

You can have access Node version on `process`.

I think [`supportTypeScript`](https://github.com/fastify/fastify-autoload/blob/69d42a56bfbbf9f2efd9ae01b8ddbfbfdd3a4ac9/lib/runtime.js#L107) should return true if node version >= 23.
Regarding testing, maybe we can create a script `unit-typescript-native-type-strippping.js` that exec ts test suite only if node version >= 23.

But we need to test if it does work first ^^.

jean-michelet:
@mcollina 
I started to work on this, I am positive we I can push a PR soon, but this error will be unreachable from Node 23:
https://github.com/fastify/fastify-autoload/blob/master/lib/find-plugins.js#L148

So it does break the test that expect this error: https://github.com/fastify/fastify-autoload/tree/master/test/issues/369

Right now, I can think about a clanky combination of conditional testing based on runtime node version and coverage ignore.

[...]

PR #440, opened 2025-01-16
test: #5555 - move to node test runner

SamSalvatico:
#### Checklist

Moving the tests from `tap` to the node test runner, based on what requested in the following issue
[Fastify #5555](https://github.com/fastify/fastify/issues/5555)

- [x] run `npm run test` and `npm run benchmark`
- [x] tests and/or benchmarks are included
- [ ] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/master/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/master/CODE_OF_CONDUCT.md)

PR #442, opened 2025-01-25
feat: support native type stripping for Node >= v23

jean-michelet:
#441

My editor has reformatted the code a bit, but eslint seems ok with some of the changes. 
I think the PR is still readable and it improves readability a bit, let me know if I should revert.

### DIFF
--- a/README.md
+++ b/README.md
@@ -319,6 +319,7 @@ provider, such as `ts-node`. Otherwise, widening the acceptance extension here w
   ```
 
 ## Override TypeScript detection using an environment variable
+> **Note**: Node >= 23 supports native type stripping, and this plugin will automatically take advantage of it. Read more about native type stripping on the [API docs](https://nodejs.org/docs/latest-v23.x/api/typescript.html#modules-typescript).
 
 It is possible to override the automatic detection of a TypeScript-capable runtime using the `FASTIFY_AUTOLOAD_TYPESCRIPT` environment variable. If set to a truthy value Autoload will load `.ts` files, expecting that node has a TypeScript-capable loader.
 
--- a/README.md
+++ b/README.md
@@ -319,7 +319,7 @@ provider, such as `ts-node`. Otherwise, widening the acceptance extension here w
   ```
 
 ## Override TypeScript detection using an environment variable
-> **Note**: Node >= 23 supports native type stripping, and this plugin will automatically take advantage of it. Read more about native type stripping on the [API docs](https://nodejs.org/docs/latest-v23.x/api/typescript.html#modules-typescript).
+> This plugin uses [native type stripping](https://nodejs.org/docs/latest-v23.x/api/typescript.html#modules-typescript) with Node 23 and later.
 
 It is possible to override the automatic detection of a TypeScript-capable runtime using the `FASTIFY_AUTOLOAD_TYPESCRIPT` environment variable. If set to a truthy value Autoload will load `.ts` files, expecting that node has a TypeScript-capable loader.
 
--- a/README.md
+++ b/README.md
@@ -319,7 +319,7 @@ provider, such as `ts-node`. Otherwise, widening the acceptance extension here w
   ```
 
 ## Override TypeScript detection using an environment variable
-> This plugin uses [native type stripping](https://nodejs.org/docs/latest-v23.x/api/typescript.html#modules-typescript) with Node 23 and later.
+This plugin uses [native type stripping](https://nodejs.org/docs/latest-v23.x/api/typescript.html#modules-typescript) with Node 23 and later.
 
 It is possible to override the automatic detection of a TypeScript-capable runtime using the `FASTIFY_AUTOLOAD_TYPESCRIPT` environment variable. If set to a truthy value Autoload will load `.ts` files, expecting that node has a TypeScript-capable loader.
 
--- a/test/commonjs/error.js
+++ b/test/commonjs/error.js
@@ -2,9 +2,12 @@
 
 const t = require('tap')
 const fastify = require('fastify')
+const runtime = require('../../lib/runtime')
+
+const typeStrippingEnabled = runtime.nodeVersion >= 23
 
 t.test('independent of module support', function (t) {
-  t.plan(8)
+  t.plan(typeStrippingEnabled ? 7 : 8)
   const app = fastify()
 
   app.register(require('./syntax-error/app'))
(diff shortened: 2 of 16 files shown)


test/commonjs/error.js
```diff
@@ -2,9 +2,12 @@
 
 const t = require('tap')
 const fastify = require('fastify')
+const runtime = require('../../lib/runtime')
+
+const typeStrippingEnabled = runtime.nodeVersion >= 23
 
 t.test('independent of module support', function (t) {
-  t.plan(8)
+  t.plan(typeStrippingEnabled ? 7 : 8)
```
jean-michelet, line 10: Can be removed when #440 is merged.

mcollina:
Can you add something in the README?

Eomm:
> My editor has reformatted the code a bit, but eslint seems ok with some of the changes.
> I think the PR is still readable and it improves readability a bit, let me know if I should revert.

If you run `npm run lint:fix`, does the change reverse itself?

jean-michelet:
> If you run npm run lint:fix, does the change reverse itself?

Not the indentation, but remove semi-colons, use single-quotes etc.

Eomm:
> > If you run npm run lint:fix, does the change reverse itself?
> 
> Not the indentation, but remove semi-colons, use single-quotes etc.

mmm so how does `npm run lint` passes? cc @voxpelli

jean-michelet:
Probably because it doesn't enforce certain formatting rules, like trailing coma for example: https://github.com/fastify/fastify-autoload/pull/442/files#diff-93d89c5bddae92725a3ce7b3a38a424edb3d4f4de3b403e34af97dadee0d583eR11

Fdawgs (changes requested):
Throwing in a block until I can properly review.

climba03003 (commented):
Too much stylistics changes.

README.md
```diff
@@ -319,6 +319,7 @@ provider, such as `ts-node`. Otherwise, widening the acceptance extension here w
   ```
 
 ## Override TypeScript detection using an environment variable
+> **Note**: Node >= 23 supports native type stripping, and this plugin will automatically take advantage of it. Read more about native type stripping on the [API docs](https://nodejs.org/docs/latest-v23.x/api/typescript.html#modules-typescript).
```
Fdawgs, line 322:
```suggestion

This plugin uses [native type stripping](https://nodejs.org/docs/latest-v23.x/api/typescript.html#modules-typescript) with Node 23 and later.
```

Fdawgs (changes requested):
As has been said, the stylistic changes need to be reverted as it's hard to see what functionality has actually changed.

Uzlopak (dismissed):
Is it still possible to use non-native-typestripping on node 23 and newer?

If not, it needs to be ensured that you can use other typescript loaders for potentially not supported features...

test/typescript-esm/app/index.ts
```diff
@@ -1,4 +1,4 @@
-import { FastifyInstance, FastifyReply, FastifyRequest } from 'fastify'
+import type { FastifyInstance, FastifyReply, FastifyRequest } from 'fastify'
```
jean-michelet, line 1: From [API docs](https://nodejs.org/docs/latest-v23.x/api/typescript.html#importing-types-without-type-keyword): 
> Due to the nature of type stripping, the type keyword is necessary to correctly strip type imports.

test/typescript-common/index.ts
```diff
@@ -0,0 +1,38 @@
+const t = require('tap')
+const fastify = require('fastify')
+
+const basicApp = require('./basic/app.ts')
```
jean-michelet, line 4: Extensions seems mandatory for ts file too.

jean-michelet:
> Is it still possible to use non-native-typestripping on node 23 and newer?

Yes:
> The flag [--no-experimental-strip-types](https://nodejs.org/docs/latest-v23.x/api/cli.html#--no-experimental-strip-types) prevents Node.js from running TypeScript files.

jean-michelet:
> it needs to be ensured that you can use other typescript loaders for potentially not supported features...

We already have tests for loaders in `scripts` folder. Is it what you were suggesting?

README.md
```diff
@@ -319,6 +319,7 @@ provider, such as `ts-node`. Otherwise, widening the acceptance extension here w
   ```
 
 ## Override TypeScript detection using an environment variable
+> This plugin uses [native type stripping](https://nodejs.org/docs/latest-v23.x/api/typescript.html#modules-typescript) with Node 23 and later.
```
Fdawgs, line 322:
```suggestion
This plugin uses [native type stripping](https://nodejs.org/docs/latest-v23.x/api/typescript.html#modules-typescript) with Node 23 and later.
```

Don't think this needs the block quote to draw attention to it. It's the first sentence of a section, redaers are going to see it.

mcollina (approved):
lgtm

test/typescript-common/index.ts
```diff
@@ -0,0 +1,38 @@
+const t = require('tap')
```
mcollina, line 1:
```suggestion
'use strict'

const t = require('tap')
```

voxpelli:
> Probably because it doesn't enforce certain formatting rules, like trailing coma for example: https://github.com/fastify/fastify-autoload/pull/442/files#diff-93d89c5bddae92725a3ce7b3a38a424edb3d4f4de3b403e34af97dadee0d583eR11

@jean-michelet @Eomm We deactivated trailing comma rules a while ago as there's no clear consensus on whether it should be enforced or banned (many existing code bases ban them, many new wants to enforce them) https://github.com/neostandard/neostandard/issues/143

Uzlopak:
is it ensured, that if you use other typescript engines, that it will use them in node 23 and not just use the type stripping?

So technically this should be possible in other engines
https://www.totaltypescript.com/erasable-syntax-only#what-does-erasablesyntaxonly-do
but fail in node23

jean-michelet:
They aren't engines but JIT-compiler TS -> JS, so Node will execute JS files when leveraged by these tools.

README.md
jean-michelet: This is not about drawing attention, just a note that is not part of the main explanations. But ok.

Fdawgs (approved):

jean-michelet:
Can you plz merge the PR if you're ok with changes @climba03003 @Uzlopak?

jean-michelet:
I merge it, so the changes can be integrated in #440

PR #452, opened 2025-03-09
refactor: leverage process.features.typescript instead of node version

jean-michelet:
Node version is not reliable enough to guess typescript support, but we can leverage `process.features.typescript`.

https://github.com/fastify/fastify-cli/pull/798#issuecomment-2708361175

### DIFF
--- a/lib/runtime.js
+++ b/lib/runtime.js
@@ -124,7 +124,7 @@ Object.defineProperties(runtime, {
   },
   supportNativeTypeScript: {
     get () {
-      cache.supportNativeTypeScript ??= ['strip', 'transform'].includes(process.features.typescript)
+      cache.supportNativeTypeScript ??= process.features.typescript !== undefined && process.features.typescript !== false
       return cache.supportNativeTypeScript
     }
   },
--- a/lib/runtime.js
+++ b/lib/runtime.js
@@ -124,7 +124,7 @@ Object.defineProperties(runtime, {
   },
   supportNativeTypeScript: {
     get () {
-      cache.supportNativeTypeScript ??= process.features.typescript !== false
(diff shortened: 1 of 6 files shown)


scripts/unit-typescript-tsx.js
```diff
@@ -1,11 +1,12 @@
 'use strict'
 
 const { exec } = require('node:child_process')
-const runtime = require('../lib/runtime')
+
+const nodeVersion = Number(process.version.split('.')[0].slice(1))
```
jean-michelet, line 5: I removed `nodeVersion` from runtime because it is only used here and prevent to reach 100% coverage on the main test suite.

test/typescript-common/index.ts
```diff
@@ -1,6 +1,6 @@
 'use script'
 
-const t = require('tap')
```
jean-michelet, line 3: Weird that it didn't fail previously :thinking:

lib/runtime.js
```diff
@@ -117,11 +117,17 @@ Object.defineProperties(runtime, {
...
         runtime.tsimp ||
-        runtime.nodeVersion >= 23
+        runtime.supportNativeTypeScript
       )
       return cache.supportTypeScript
     }
   },
+  supportNativeTypeScript: {
+    get () {
+      cache.supportNativeTypeScript ??= ['strip', 'transform'].includes(process.features.typescript)
```
jean-michelet, line 127: Should it be?
```ts
process.features.typescript !== undefined && process.features.typescript !== false
```

mcollina: yes!

mcollina:
looks great, just one thing.

mcollina (approved):
lgtm

PR #454, opened 2025-03-27
Resolve dirNameRoutePrefix interferes with auto-hooks

### LINKED ISSUE #453: dirNameRoutePrefix interferes with auto-hooks
### Prerequisites

- [x] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

5.2.1

### Plugin version

6.1.0

### Node.js version

20.12.2

### Operating system

macOS

14.4 (23E214)

### Description

I've encountered an issue when using the dirNameRoutePrefix option together with autoHooks.

When dirNameRoutePrefix is set to false, auto-hooks are applied not only to the parent folder but also to its first-level child routes. Additionally, if dirNameRoutePrefix is a function that selectively returns false for specific folders, only the parent folder’s routes will use the hooks from the folder marked as false.

Example Configurations
1. Global false value (Issue)
```
fastify.register(pluginAutoload, {
  ...defaultAutoloadOptions,
  // ISSUE: The parent folder routes and the first-level child routes 
  // will inherit the last registered hook (e.g., third.hook.mjs).
  dirNameRoutePrefix: false,
})
```
2. Conditional Prefixing (Issue)
```
fastify.register(pluginAutoload, {
  ...defaultAutoloadOptions,
  dirNameRoutePrefix: (folderParent, folderName) => {
    if (folderName === 'third') {
      return 'yellow-submarine';
    }
    // ISSUE: The parent folder routes and the first child routes 
    // will still inherit the last registered hook (e.g., third.hook.mjs).
    return false;
  },
})
```
3. More Granular Control (Partial Issue)
```
fastify.register(pluginAutoload, {
  ...defaultAutoloadOptions,
  dirNameRoutePrefix: function rewrite(folderParent, folderName) {
    if (folderName === 'third') {
      return 'yellow-submarine'; // This works fine.
    }
    if (folderName === 'first') {
      return false; 
      // ISSUE: In this case, the parent folder will still use hooks 
      // from inside this folder.
    }
    return folderName; // This works fine.
  },
})
```

Additional Notes
In my actual application (which I can't share), I use the second example because I want most routes to have no prefix while only a few specific ones are prefixed. However, in this setup, I've encountered cases where the hook inside a falsy folder is not applied at all. This behavior is inconsistent, and I haven't been able to reproduce it in my test repository.

https://github.com/MrWook/reproduction-fastify-autoload-hook-issue

### Expected Behavior

dirNameRoutePrefix is not interfering with autohooks so that the hooks are still only used in the specific folder

MrWook:
Should resolve https://github.com/fastify/fastify-autoload/issues/453

Previously the prefix from `dirNameRoutePrefix` was used as a key to build the plugins which resulted in an unexpected behaviour.
For example with:
```
fastify.register(pluginAutoload, {
  autoHooks: true,
  cascadeHooks: true,
  // ISSUE: The parent folder routes and the first-level child routes will inherit the last registered hook (e.g., third.hook.mjs).
  dirNameRoutePrefix: false,
})
```
The routes in the parent folder were using the child folder hooks and if you have multiple child folder in the parent folder all of them would be using the last registered hook [see the issue](https://github.com/fastify/fastify-autoload/issues/453).

I tested this in my reproduction [repository](https://github.com/MrWook/reproduction-fastify-autoload-hook-issue) and it seems to be working as expected now.

### DIFF
--- a/lib/find-plugins.js
+++ b/lib/find-plugins.js
@@ -8,7 +8,7 @@ async function findPlugins (dir, options) {
   const { opts, prefix } = options
 
   const pluginTree = {
-    [prefix || '/']: { hooks: [], plugins: [] }
+    [dir]: { hooks: [], plugins: [] }
   }
 
   await buildTree(pluginTree, dir, { prefix, opts, depth: 0, hooks: [] })
@@ -18,8 +18,8 @@ async function findPlugins (dir, options) {
 
 async function buildTree (pluginTree, dir, { prefix, opts, depth, hooks }) {
   // check to see if hooks or plugins have been added to this prefix, initialize if not
-  if (!pluginTree[prefix]) {
-    pluginTree[prefix] = { hooks: [], plugins: [] }
+  if (!pluginTree[dir]) {
+    pluginTree[dir] = { hooks: [], plugins: [] }
   }
 
   const dirEntries = await readdir(dir, { withFileTypes: true })
@@ -47,6 +47,7 @@ function findCurrentDirHooks (pluginTree, { dir, dirEntries, hooks, opts, prefix
   let currentDirHooks = []
   // Hooks were passed in, create new array specific to this plugin item
   for (const hook of hooks) {
+    console.log(hook)
     currentDirHooks.push(hook)
   }
 
@@ -65,7 +66,7 @@ function findCurrentDirHooks (pluginTree, { dir, dirEntries, hooks, opts, prefix
     currentDirHooks.push({ file, type })
   }
 
-  pluginTree[prefix || '/'].hooks = currentDirHooks
+  pluginTree[dir].hooks = currentDirHooks
 
   return currentDirHooks
 }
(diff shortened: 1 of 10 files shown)


lib/find-plugins.js
```diff
@@ -47,6 +47,7 @@ function findCurrentDirHooks (pluginTree, { dir, dirEntries, hooks, opts, prefix
   let currentDirHooks = []
   // Hooks were passed in, create new array specific to this plugin item
   for (const hook of hooks) {
+    console.log(hook)
```
jean-michelet, line 50:
```suggestion
```

jean-michelet (commented):
Can you plz write a test that proove the bug is fixed: not having hooks registered in children plugins when `dirNameRoutePrefix === false` or `dirNameRoutePrefix === () => false.`.

You can write a test [here](https://github.com/fastify/fastify-autoload/tree/main/test/issues), the folder should have the name of the issue.

MrWook:
I added tests and verified them against the previous implementation, which failed as I described earlier. These tests now effectively cover the issue.

jean-michelet (approved):

mcollina (approved):
lgtm

PR #463, opened 2025-05-20
Fixing files imported by module got bypass through isPluginOrModule check

### LINKED ISSUE #462: Autoload a file that does not export as a plugin cause vague error message.
### Prerequisites

- [x] I have searched existing issues to ensure the issue has not already been raised

### Issue

Here is my autoload plugin setup
```ts
 fastify
  .register(autoload, {
    dir: folder,
    forceESM: true,
    matchFilter: (path) => {
      const isMatched =
        path.includes('.api.') ||
        path.includes('.route.') ||
        path.includes('.graphql.');
      if (opts.debug && isMatched) {
        Logger.info(path);
      }
      return isMatched;
    },
    dirNameRoutePrefix: false,
    logLevel: 'debug',
    options: {
      prefix: '/api',
    },
  })
```

It autoloads all the file with `.api.` in the folder I want. But the problem is when I create a new blank file, or editting a previous working file into a non-plugin file by accident, it causes the following autoload error where it found a file that does not export a plugin function. 

The thing that bugs me is that I have no idea which file that flags the error, this is really vague and always made me search through my changes to see which one was causing problems, or straight up have to `console.log` in the matchFilter function to see which one was last checked. Can we have a more explicit error message here?

```bash
/Users/link/Documents/Repo/BC/bluecheck-be/node_modules/.pnpm/avvio@9.1.0/node_modules/avvio/lib/validate-plugin.js:19
      throw new AVV_ERR_PLUGIN_NOT_VALID(typeof maybePlugin)
            ^
FastifyError [Error]: Plugin must be a function or a promise. Received: 'object'
    at validatePlugin (/Users/link/Documents/Repo/BC/bluecheck-be/node_modules/.pnpm/avvio@9.1.0/node_modules/avvio/lib/validate-plugin.js:19:13)
    at Boot._addPlugin (/Users/link/Documents/Repo/BC/bluecheck-be/node_modules/.pnpm/avvio@9.1.0/node_modules/avvio/boot.js:166:3)
    at Boot.use (/Users/link/Documents/Repo/BC/bluecheck-be/node_modules/.pnpm/avvio@9.1.0/node_modules/avvio/boot.js:139:25)
    at server.<computed> [as register] (/Users/link/Documents/Repo/BC/bluecheck-be/node_modules/.pnpm/avvio@9.1.0/node_modules/avvio/boot.js:213:14)
    at registerPlugin (/Users/link/Documents/Repo/BC/bluecheck-be/node_modules/.pnpm/@fastify+autoload@6.3.0/node_modules/@fastify/autoload/index.js:181:11)
    at registerAllPlugins (/Users/link/Documents/Repo/BC/bluecheck-be/node_modules/.pnpm/@fastify+autoload@6.3.0/node_modules/@fastify/autoload/index.js:157:17)
    at registerNode (/Users/link/Documents/Repo/BC/bluecheck-be/node_modules/.pnpm/@fastify+autoload@6.3.0/node_modules/@fastify/autoload/index.js:135:5)
    at loadPlugins (/Users/link/Documents/Repo/BC/bluecheck-be/node_modules/.pnpm/@fastify+autoload@6.3.0/node_modules/@fastify/autoload/index.js:64:5)
    at async autoload (/Users/link/Documents/Repo/BC/bluecheck-be/node_modules/.pnpm/@fastify+autoload@6.3.0/node_modules/@fastify/autoload/index.js:22:3) {
  code: 'AVV_ERR_PLUGIN_NOT_VALID',
  statusCode: 500
}
```

linkthai:
Pull request for issue https://github.com/fastify/fastify-autoload/issues/462

So in the issue above, I reported that the empty files got thrown a vague error, but turns out these files are supposed to be ignored. And I found out that this only happens in files imported by module as I have the `forceESM: true`.

When a file is imported, the `input` and `inputType` would looks like this
```ts 
// For empty.cjs and empty.js file 
[Module: null prototype] { default: {} } // input
[object Module] // inputType

// For empty.mjs file
[Module: null prototype] {  } // input
[object Module] // inputType

// For api.js file that export a proper module
[Module: null prototype] { default: { default: [Getter] } } // input
[object Module] // inputType

// For object.js file that export an incorrect module
[Module: null prototype] { default: { default: [Getter] } } // input 
[object Module] // inputType

// For hello.js file that does NOT export a module
[Module: null prototype] { default: { hello: [Getter] } } // input
[object Module] // inputType
```
All these scenarios would pass the `isPluginOrModule` check since the `inputType` is `Module` even though they should not.

I fix it by providing an extra flag `importedAsModule` whenever a file is imported. If a module got imported they are always wrapped by a `{ default: {} }` (except the mjs file), a cjs file with a default function would have its `input` looks like this `{ default: { default: [AsyncFunction] } }` so we need to cast the `inputType` through `input.default` instead. 

In all the case except mjs, `inputType` would be an object instead of module, thus will run into the second if condition and loop back into the correct `inputType` check. For mjs files, it would run into the correct check on the first loop.
```ts
function isPluginOrModule (input, importedAsModule) {
  let result = false

  const inputType = Object.prototype.toString.call(importedAsModule ? input.default : input)
  if (pluginOrModulePattern.test(inputType) === true) {
    result = true
  } else if (Object.hasOwn(input, 'default')) {
    result = isPluginOrModule(input.default)
  } else {
    result = isRouteObject(input)
  }

  return result
}
```

#### Checklist

- [ ] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/main/CODE_OF_CONDUCT.md)


index.js
```diff
@@ -67,14 +67,15 @@ async function loadPlugins ({ pluginTree, options, opts, fastify }) {
...
+  let content, importedAsModule
   if (forceESM || type === 'module' || runtime.forceESM) {
     content = await import(pathToFileURL(file).href)
+    importedAsModule = true
   } else {
     content = require(file)
   }
 
-  if (isPluginOrModule(content) === false) {
+  if (isPluginOrModule(content, importedAsModule) === false) {
```
Uzlopak, line 78: actually... ist the following not enough?

```suggestion
  if (isPluginOrModule(content.default || content) === false) {
```

of maybe

```suggestion
  if (isPluginOrModule(importedAsModule ? content.default : content, importedAsModule) === false) {
```

Then you dont need any changes in `isPluginOrModule`

linkthai: thank you, I'd go with the second because it's clearer. I was too tunneling on fixing the function instead of the parameter.

test/issues/462/test.js
```diff
@@ -0,0 +1,30 @@
+'use strict'
+
+const { after, before, describe, it } = require('node:test')
+const path = require('node:path')
+const Fastify = require('fastify')
+const autoLoad = require('../../../')
+const assert = require('node:assert')
+
+describe('Issue 462 tests', function () {
```
Uzlopak, line 9: A more descriptive title would be better


```suggestion
describe('ignore invalid and/or empty .js and .mjs files when using esm', function () {
```

Uzlopak (changes requested):
Just some minor request

test/issues/462/test.js
jean-michelet: That's fine I think, if we need more context, we can read the issue.
Eventually we can open a PR to update the existing `describe` test files (which aren't descriptive either).

linkthai: I added a more descriptive title for the test case
```ts
describe('Issue 462: ignore invalid and/or empty .js and .mjs files when using esm', function () {
```

Uzlopak (approved):
LGTM

jean-michelet (approved):

index.js
```diff
@@ -67,14 +67,15 @@ async function loadPlugins ({ pluginTree, options, opts, fastify }) {
...
   let content
-  if (forceESM || type === 'module' || runtime.forceESM) {
+  if (importedAsModule) {
     content = await import(pathToFileURL(file).href)
   } else {
     content = require(file)
   }
 
-  if (isPluginOrModule(content) === false) {
+  if (isPluginOrModule(importedAsModule ? content.default : content) === false) {
```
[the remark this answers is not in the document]
jean-michelet: Let @climba03003 merge when ok with this.

climba03003 (approved):

index.js
```diff
@@ -67,18 +67,20 @@ async function loadPlugins ({ pluginTree, options, opts, fastify }) {
...
     // We must have something that resembles a Fastify plugin function, or that
     // can be converted into one, that can eventually be passed to `avvio`. If
     // it is anything else, skip automatic loading of this item.
     log.debug({ file }, 'skipping autoloading of file because it does not export a Fastify plugin compatible shape')
+    console.log('hereer')
```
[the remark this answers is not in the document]
linkthai: haha sorry that was embarrassing, I totally forgot about it

PR #471, opened 2025-07-08
docs(readme): use h3 headers for options

### LINKED ISSUE #465: Add support for dynamic route segments
### Prerequisites

- [x] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

This library should support loading dynamic route segments. In a route like '/users/:id/posts' there is no way to put the ':id' segment into the directory structure. I would like to put these route inside of '/users/[id]/posts/index.js', but there isn't any way to do that right now.

### Motivation

Suppose that I have a module for users as follows:

```
.
└── routes
    └── users
        └── index.js
```

```
export default async function () => {
  fastify.get('/', (req, reply) => {
    reply.send('/users');
  });

  fastify.get('/:id', (req, reply) => {
    reply.send('/users/${req.params.id}');
  });
}
```

This app has two routes right now, one gets all the users, and another gets a specific user by an ID.

Now suppose that I want to create more routes, routes to add a user, routes to update a user, and this index.js file starts getting full of routes for modifying users. What should I do if I want to now add a nested resource? Imagine I want to add blog posts, so each user can have a list of blog posts. I might add a route like GET '/users/:id/posts' or PATCH '/users/:id/posts/:id'.

This module is going to get quite cluttered if I put it all here.

Really what I want is to be able to do this with directories instead of fastify.get. I want to be able to have a folder name like [id] map to :id in the route segments. NextJS does things like this already, which is why I expected it to be a feature of this library.

### Example

```
.
└── routes
    └── users
        ├── [id]
        │   └── posts
        │       └── index.js
        └── index.js
```

Ben-at-Catalyst:
I wasn't able to find the `routeParams` option on the README, I had been searching for "dynamic", "route segment", etc. I opened an issue and was informed of the `routeParams` option.

This PR should help future users search for that option as I did.

@jean-michelet Also suggested that I change the options to use h4 headings instead of plain list entries, so that they can be linked, so I've done that here as well.

Closes #465

#### Checklist

- [X] run `npm run test && npm run benchmark --if-present`
- [X] documentation is changed or added
- [X] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)
      and the [Code of conduct](https://github.com/fastify/.github/blob/main/CODE_OF_CONDUCT.md)

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ Folder structure:
 
 Autoload can be customized using the following options:
 
-#### `dir` (required) - Base directory containing plugins to be loaded
+### `dir` (required) - Base directory containing plugins to be loaded
 
   Each script file within a directory is treated as a plugin unless the directory contains an index file (e.g. `index.js`). In which case, only the index file (and the potential sub-directories) will be loaded.
 
@@ -101,7 +101,7 @@ Autoload can be customized using the following options:
   - `.mjs` (ES modules)
   - `.ts` (TypeScript)
 
-#### `dirNameRoutePrefix` (optional) - Default: true. Determines whether routes will be automatically prefixed with the subdirectory name in an autoloaded directory. It can be a sync function that must return a string that will be used as prefix, or it must return `false` to skip the prefix for the directory.
+### `dirNameRoutePrefix` (optional) - Default: true. Determines whether routes will be automatically prefixed with the subdirectory name in an autoloaded directory. It can be a sync function that must return a string that will be used as prefix, or it must return `false` to skip the prefix for the directory.
 
   ```js
   fastify.register(autoLoad, {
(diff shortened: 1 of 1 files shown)


jean-michelet (commented):
I confirm I suggested we use h4 (or another header) so options can be linked.
I suggest to wait for @Fdawgs feedback before continuing on this PR.

Fdawgs (changes requested):
Good idea, I'd go with h3 though to align with the main Fastify repo's style.

Ben-at-Catalyst:
@jean-michelet and @Fdawgs, I have made the requested changes. Descriptions are now on new-lines and I have used h3 instead of h4.

jean-michelet (approved):

gurgunday (approved):
lgtm

jean-michelet:
Can you plz merge if you agree @Fdawgs?

Fdawgs (changes requested):
Thanks @Ben-at-Catalyst, one last thing, could you sort the indentation please? The indentation of the text and js blocks following these new headers is no longer needed.

Was obviously needed when they were bullet points.

gurgunday (approved):
_lgtm_

Fdawgs (approved):

PR #497, opened 2026-03-08
fix: scope autohooks setNotFoundHandler under prefixed autoload

### LINKED ISSUE #326: setNotFoundHandler is not encapsulated when called inside autohooks.js

SergioChan:
## Summary
- scope `autohooks` registration under a shared plugin prefix when all plugins in the node resolve to the same prefix
- strip that shared prefix from child plugin registrations inside the scoped context so routes are not double-prefixed
- add a regression test for issue #326 proving `setNotFoundHandler` from `autohooks.js` only affects `/api/*` requests

## Testing
- `npm run unit test/issues/326/test.js test/issues/374/test.js test/issues/376/test.js test/issues/453/test.js`

## Related
Fixes #326

### DIFF
--- a/index.js
+++ b/index.js
@@ -135,6 +135,10 @@ function registerNode (node, fastify) {
   if (node.hooks.length === 0) {
     registerAllPlugins(fastify, node)
   } else {
+    const hooksPrefix = findCommonHooksPrefix(node)
+    const scopedPluginsMeta = hooksPrefix ? buildScopedPluginsMeta(node.pluginsMeta, hooksPrefix) : node.pluginsMeta
+    const scopedNode = hooksPrefix ? { ...node, pluginsMeta: scopedPluginsMeta } : node
+
     const composedPlugin = async function (app) {
       // find hook functions for this prefix
       for (const hookFile of node.hooks) {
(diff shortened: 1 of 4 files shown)


mcollina (approved):
lgtm

gurgunday (approved):

PR #516, opened 2026-06-12
fix: match ignorePattern against relative path instead of entry basename

LeSingh1:
When ignorePattern contains a path separator (e.g. /private\/plugin/ to
skip routes/private/plugin.js without affecting routes/public/plugin.js),
it never matched because the check tested the bare directory or file name
rather than the path relative to opts.dir.

matchFilter and ignoreFilter already receive the full relative path with
forward-slash normalisation. This PR applies the same approach to the
ignorePattern check so the two APIs behave consistently.

The change is one line larger than the original: compute the relative
path from opts.dir with the same backslash normalisation used in
accumulatePlugin, then test the pattern against that instead of
dirEntry.name. Existing tests that pass a basename-only pattern (e.g.
/^ignored/) keep working because the relative path still ends with the
basename, so basename-only patterns still match.

A new test in test/issues/519/ covers the failing case: two sibling
subdirectories both containing a plugin.js, ignorePattern set to
/private\/plugin/, verifying that only the private route is excluded.

All 272 unit tests pass, lint is clean, and code coverage remains at 100%.

### DIFF
--- a/lib/find-plugins.js
+++ b/lib/find-plugins.js
@@ -87,8 +87,11 @@ function processIndexDirEntryIfExists (pluginTree, { opts, dirEntries, dir, pref
 
 async function processDirContents (pluginTree, { dirEntries, opts, indexDirEntry, prefix, dir, depth, currentDirHooks }) {
   for (const dirEntry of dirEntries) {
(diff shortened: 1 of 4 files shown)


mcollina (approved):
lgtm

Tony133 (approved):

